PR #4, opened 2019-05-31
Enable preview from dev branch builds

MIchaelMainer:
**previewBuild.yml**
Specifies a build pipeline that targets the latest commit to the dev branch. This pipeline will not run on PR. This pipeline performs the following
* Makes sure that the version number has been incremented. 
* Builds the solution and runs the tests.
* Signing
* Sets or increments the prerelease version
* Publishes the built NuGet package so that it can be picked up by release pipeline.

**EnableSigning.ps1**
Sets the csproj ready for signing.

**IncrementPreviewVersion.ps1**
Increments the preview version based on whether there is a publicly published preview version. Does nothing if the VersionSuffix is set in the csproj.


MIchaelMainer:
Signing is not working as expected.

https://o365exchange.visualstudio.com/O365%20Sandbox/_build/results?buildId=1351334&view=results

PR #11, opened 2019-06-12
Update from dotnet pack to msbuild pack

MIchaelMainer:
Xamarin update requires us to use msbuild pack to create a NuGet
package.

On branch mm/supportXamarinPack
Changes to be committed:
    modified:   pipelines/previewBuild.yml
    modified:   pipelines/productionBuild.yml


MIchaelMainer:
Strange. The two pipeline in the checks were updated, but are not supposed to run on a PR.

PR #14, opened 2019-06-21
Updating Retry Handler to use time limit instead of count limit

irvinesunday:
Fixes https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/441

- Maximum request retries will be based against the maximum retries time limit specified by the client, or the default maximum retries time limit value.
- If the the delay is greater than the maximum retries time limit specified, then there will be no request retry and the response will be returned. 

### Other links
- Link to the Retry Handler design spec: https://github.com/microsoftgraph/msgraph-sdk-design/pull/18

### DIFF
--- a/src/Microsoft.Graph.Core/Exceptions/ErrorConstants.cs
+++ b/src/Microsoft.Graph.Core/Exceptions/ErrorConstants.cs
@@ -20,8 +20,6 @@ namespace Microsoft.Graph
 
             internal static string TooManyRedirects = "tooManyRedirects";
 
-            internal static string TooManyRetries = "tooManyRetries";
-
             internal static string MaximumValueExceeded = "MaximumValueExceeded";
 
             internal static string InvalidArgument = "invalidArgument";
@@ -46,8 +44,6 @@ namespace Microsoft.Graph
             internal static string RequestUrlMissing = "Request URL is required to send a request.";
 
             internal static string TooManyRedirectsFormatString = "More than {0} redirects encountered while sending the request.";
-
-            internal static string TooManyRetriesFormatString = "More than {0} retries encountered while sending the request.";
 
             internal static string UnableToCreateInstanceOfTypeFormatString = "Unable to create an instance of type {0}.";
 
--- a/src/Microsoft.Graph.Core/Exceptions/ErrorConstants.cs
+++ b/src/Microsoft.Graph.Core/Exceptions/ErrorConstants.cs
@@ -25,6 +25,8 @@ namespace Microsoft.Graph
             internal static string MaximumValueExceeded = "MaximumValueExceeded";
 
             internal static string InvalidArgument = "invalidArgument";
+
+            internal static string MaximumRetryTimeReached = "maximumRetryTimeReached";
         }
 
         internal static class Messages
@@ -64,6 +66,8 @@ namespace Microsoft.Graph
             internal static string InvalidDependsOnRequestId = "Corresponding batch request id not found for the specified dependsOn relation.";
 
             public static string InvalidProxyArgument = "Proxy cannot be set more once. Proxy can only be set on the proxy or defaultHttpHandler argument and not both.";
+
+            internal static string MaximumRetryTimeReached = "Maximum time for request retries reached.";
         }
     }
 }
--- a/src/Microsoft.Graph.Core/Exceptions/ErrorConstants.cs
+++ b/src/Microsoft.Graph.Core/Exceptions/ErrorConstants.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Graph
 
             internal static string InvalidArgument = "invalidArgument";
 
-            internal static string MaximumRetryTimeReached = "maximumRetryTimeReached";
+            internal static string MaximumRetriesTimeLimitReached = "maximumRetriesTimeLimitReached";
         }
 
         internal static class Messages
@@ -67,7 +67,7 @@ namespace Microsoft.Graph
 
             public static string InvalidProxyArgument = "Proxy cannot be set more once. Proxy can only be set on the proxy or defaultHttpHandler argument and not both.";
 
-            internal static string MaximumRetryTimeReached = "Maximum time for request retries reached.";
+            internal static string MaximumRetriesTimeLimitReached = "Maximum time of {0} for request retries reached.";
         }
     }
 }
--- a/src/Microsoft.Graph.Core/Exceptions/ErrorConstants.cs
+++ b/src/Microsoft.Graph.Core/Exceptions/ErrorConstants.cs
@@ -25,8 +25,6 @@ namespace Microsoft.Graph
             internal static string MaximumValueExceeded = "MaximumValueExceeded";
 
             internal static string InvalidArgument = "invalidArgument";
-
-            internal static string MaximumRetriesTimeLimitReached = "maximumRetriesTimeLimitReached";
         }
 
         internal static class Messages
@@ -66,8 +64,6 @@ namespace Microsoft.Graph
             internal static string InvalidDependsOnRequestId = "Corresponding batch request id not found for the specified dependsOn relation.";
 
             public static string InvalidProxyArgument = "Proxy cannot be set more once. Proxy can only be set on the proxy or defaultHttpHandler argument and not both.";
-
-            internal static string MaximumRetriesTimeLimitReached = "Maximum time of {0} for request retries reached.";
         }
     }
 }
--- a/src/Microsoft.Graph.Core/Exceptions/ErrorConstants.cs
+++ b/src/Microsoft.Graph.Core/Exceptions/ErrorConstants.cs
@@ -20,6 +20,8 @@ namespace Microsoft.Graph
 
             internal static string TooManyRedirects = "tooManyRedirects";
 
+            internal static string TooManyRetries = "tooManyRetries";
+
             internal static string MaximumValueExceeded = "MaximumValueExceeded";
 
             internal static string InvalidArgument = "invalidArgument";
@@ -44,6 +46,8 @@ namespace Microsoft.Graph
             internal static string RequestUrlMissing = "Request URL is required to send a request.";
 
             internal static string TooManyRedirectsFormatString = "More than {0} redirects encountered while sending the request.";
+
+            internal static string TooManyRetriesFormatString = "More than {0} retries encountered while sending the request.";
 
             internal static string UnableToCreateInstanceOfTypeFormatString = "Unable to create an instance of type {0}.";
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.Graph.Core/Extensions/BaseRequestExtensions.cs
```diff
@@ -107,6 +107,29 @@ public static class BaseRequestExtensions
...
         }
 
+        /// <summary>
+        /// Sets the maximum time for request retries to the default Retry Middleware Handler for this request.
+        /// This only works with the default Retry Middleware Handler.
+        /// If you use a custom Retry Middleware Handler, you have to handle it's retrieval in your implementation.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="baseRequest">The <see cref="BaseRequest"/> for the request.</param>
+        /// <param name="maxRetryTime">The maxRetryTime for the request.</param>
```
MIchaelMainer: Did you intend to rename this retriesTimeLimit? Please address throughout.

src/Microsoft.Graph.Core/Requests/Middleware/RetryHandler.cs
```diff
@@ -58,7 +58,7 @@ protected async override Task<HttpResponseMessage> SendAsync(HttpRequestMessage
 
             var response = await base.SendAsync(httpRequest, cancellationToken);
 
-            if(ShouldRetry(response) && httpRequest.IsBuffered() && RetryOption.MaxRetry > 0 && RetryOption.ShouldRetry(RetryOption.Delay, 0, response))
+            if(ShouldRetry(response) && httpRequest.IsBuffered() && RetryOption.RetriesTimeLimit > 0 && RetryOption.ShouldRetry(RetryOption.Delay, 0, response))
```
MIchaelMainer: This may change behavior, although I'm not certain yet that this is a negative thing since this is retry, and I can't off the top of my head figure a way that a customer could peg to the behavior. 

We should default to use MaxRetry and only use RetriesTimeLimit if the customer specifies it. I'd like to hear folks thoughts on this. This will affect the review on parts of this PR.

Also, please add comments to describe what this long condition is doing, for our future us.

src/Microsoft.Graph.Core/Extensions/BaseRequestExtensions.cs
irvinesunday: Yes I do. Change duly effected. Thanks

src/Microsoft.Graph.Core/Extensions/BaseRequestExtensions.cs
```diff
@@ -107,6 +107,29 @@ public static class BaseRequestExtensions
...
+        /// <summary>
+        /// Sets the maximum time for request retries to the default Retry Middleware Handler for this request.
+        /// This only works with the default Retry Middleware Handler.
+        /// If you use a custom Retry Middleware Handler, you have to handle it's retrieval in your implementation.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="baseRequest">The <see cref="BaseRequest"/> for the request.</param>
+        /// <param name="retriesTimeLimit">The retriestimelimit for the request.</param>
+        /// <returns></returns>
+        public static T WithRetriesTimeLimit<T>(this T baseRequest, int retriesTimeLimit) where T : IBaseRequest
```
darrelmiller: I suggest we make this an overload of WithMaxRetry and use Timespan as the type.  Using int as the datatype makes it unclear when reading code if this is milliseconds, seconds, minutes.

src/Microsoft.Graph.Core/Exceptions/ErrorConstants.cs
```diff
@@ -20,11 +20,11 @@ internal static class Codes
 
             internal static string TooManyRedirects = "tooManyRedirects";
```
darrelmiller: Why have we removed this error?

irvinesunday: Since we are moving away from count-based retry handling to time-based, I saw it appropriate to have it replaced with:
https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/7b8c7644080acb4afb556875693421cbb0dcf782/src/Microsoft.Graph.Core/Exceptions/ErrorConstants.cs#L27

and

https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/7b8c7644080acb4afb556875693421cbb0dcf782/src/Microsoft.Graph.Core/Exceptions/ErrorConstants.cs#L66

darrelmiller: It wasn't my understanding that we were moving away from retry count based handling, we were simply adding another approach.  For operations that potentially take a varying amount of time to execute, time based retry may not be appropriate.

src/Microsoft.Graph.Core/Extensions/BaseRequestExtensions.cs
irvinesunday: I could make it an overload method of `WithMaxRetry`. 
With regards to the datatype, I chose to use `int` so as to maintain consistency within the project; The `Delay` parameter is `int` typed as well. I could probably add more descriptive comments in the xml documentation.

darrelmiller: Being consistent with delay make sense considering Delay() is public.  But I think that Delay() should not be public.  @peombwa Any idea why we made Delay public?  I think anyone looking at the surface API of this class would question why we chose int.  I definitely think we should make retriesTimeLimit a TimeSpan.

src/Microsoft.Graph.Core/Exceptions/ErrorConstants.cs
irvinesunday: @darrelmiller so this would ideally mean having an overloaded method for this: 
https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/2c23a8bdeeaa608e7e5e36494b443e5982fbef6b/src/Microsoft.Graph.Core/Requests/Middleware/RetryHandler.cs#L75-L114

1) for the **count-based** retry task, and
2) for the **retry-based** retry task

and check which option to use here:
https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/2c23a8bdeeaa608e7e5e36494b443e5982fbef6b/src/Microsoft.Graph.Core/Requests/Middleware/RetryHandler.cs#L61-L64

Ideas are welcomed @peombwa

src/Microsoft.Graph.Core/Extensions/BaseRequestExtensions.cs
irvinesunday: I have chosen to go with TimeSpan :)

peombwa: It was inherited code, but since we've already changed its interface in this PR, we can go ahead and change it to internal (our tests depends on it). I'm all for TimeSpan as opposed to int.

src/Microsoft.Graph.Core/Exceptions/ErrorConstants.cs
darrelmiller: That would create a lot of redundancy.  I would try and add a condition to the while loop.  The max retry condition should stay and we should add another condition that effectively says "if someone has specified a retriesTimeLimit and we won't exceed the time limit with with this retry then let's do it".

irvinesunday: @darrelmiller in essence you are saying that the count-based retry condition should be run side-by-side with the time-based retry condition? Won't one cancel out the other in certain scenarios before success? e.g. 3 retries of 1 sec. where a user specified a retriesTimeLimit of 10 sec. would probably receive a failure before reaching the 10 sec. time limit? or 3 retries of 10 sec. where a user specified a retriesTimeLimit of 5 sec. would also expect a failure before counting down the full 3 retries?

MIchaelMainer: Which ever limit is hit first should be respected. So, with your example, and assuming a max retry count of 3, the retry count limit would be hit before the max cumulative time limit.

src/Microsoft.Graph.Core/Extensions/BaseRequestExtensions.cs
irvinesunday: Changing the `Delay()` to TimeSpan might lead to breaking changes as we pass the `Delay()` value as an `int` to this delegate: 
https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/2c23a8bdeeaa608e7e5e36494b443e5982fbef6b/src/Microsoft.Graph.Core/Requests/Middleware/Options/RetryHandlerOption.cs#L79

src/Microsoft.Graph.Core/Requests/Middleware/RetryHandler.cs
```diff
@@ -200,13 +150,12 @@ private void AddOrUpdateRetryAttempt(HttpRequestMessage request, int retry_count
         /// <param name="retry_count">The retry counts</param>
         /// <param name="delay">Delay value in seconds.</param>
         /// <param name="delayInSeconds"></param>
-        /// <param name="cumulativeDelay">The cumulative retry time </param>
         /// <param name="cancellationToken">The cancellationToken for the Http request</param>
         /// <returns>The <see cref="Task"/> for delay operation.</returns>
-        public Task Delay(HttpResponseMessage response, int retry_count, int delay, out double delayInSeconds, ref TimeSpan cumulativeDelay, CancellationToken cancellationToken)
+        public Task Delay(HttpResponseMessage response, int retry_count, int delay, out double delayInSeconds, CancellationToken cancellationToken)
```
MIchaelMainer: This shouldn't be `public`. This should be at least `internal`, ideally `private`.

irvinesunday: Using `internal` as this is accessed from one of the test class methods.
https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/2c23a8bdeeaa608e7e5e36494b443e5982fbef6b/tests/Microsoft.Graph.DotnetCore.Core.Test/Requests/Middleware/RetryHandlerTests.cs#L276-L284

src/Microsoft.Graph.Core/Extensions/BaseRequestExtensions.cs
MIchaelMainer: It would be a breaking change to change Delay() to use TimeSpan, but then so is making it internal. It wasn't intended to be public so this is really a bug fix. Go ahead and make this change.

PR #37, opened 2019-09-11
Improve ServiceException/Error class information

### LINKED ISSUE #23: Developers get access to all actionable information in exception classes
This is a high level summary issue to keep track of work that needs to be done on the Exception classes in the core library.

- [x] In some scenarios some error response information is lost. https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/316
- [x] ClientRequestId should be exposed as a first class property on the Error object.
- [x] Support full [error response body](http://docs.oasis-open.org/odata/odata/v4.01/csprd05/part1-protocol/odata-v4.01-csprd05-part1-protocol.html#_Toc14172757).

MIchaelMainer:
**modified:   src/Microsoft.Graph.Core/Exceptions/Error.cs**

Added Target, Details, and ClientRequestId properties and updated the
ToString() method.

**new file:   src/Microsoft.Graph.Core/Exceptions/ErrorDetails.cs**

Supports the Error.Details property per OData spec, 9.4 Error Response Body details object.
http://docs.oasis-open.org/odata/odata/v4.01/csprd05/part1-protocol/odata-v4.01-csprd05-part1-protocol.html#_Toc14172757

**modified:   src/Microsoft.Graph.Core/Exceptions/ServiceException.cs**

Added support for adding the raw response body.

**modified:   src/Microsoft.Graph.Core/Requests/HttpProvider.cs**

Added support for adding the raw response body and ClientRequestId for
ServiceExceptions.

modified:   tests/Microsoft.Graph.DotnetCore.Core.Test/Exceptions/ErrorTests.cs
modified:   tests/Microsoft.Graph.DotnetCore.Core.Test/Requests/HttpProviderTests.cs

**Addresses**

* Fixes #23
* Fixes microsoftgraph/msgraph-sdk-dotnet#316


src/Microsoft.Graph.Core/Exceptions/Error.cs
```diff
@@ -12,15 +12,35 @@ namespace Microsoft.Graph
...
     /// The error object that handles unsuccessful responses returned from the service.
+    /// Models OData protocol, 9.4 Error Response Body
+    /// http://docs.oasis-open.org/odata/odata/v4.01/csprd05/part1-protocol/odata-v4.01-csprd05-part1-protocol.html#_Toc14172757
     /// </summary>
     [JsonObject(MemberSerialization = MemberSerialization.OptIn)]
     public class Error
     {
         /// <summary>
-        /// The HTTP status code.
+        /// This code serves as a sub-status for the HTTP error code specified in the response.
```
darrelmiller: The code property for the Error object at the root of the document is a 1:1 mapping to the HTTP status code.  When the Error object is in InnerError or ErrorDetails then the code becomes a sub-code.

darrelmiller: We should sync with Saurabh as they are currently making changes to ODataLib in this area and we should try and align.  /cc @madansr7

MIchaelMainer: According to spec, it is supposed to a sub status code:

*code*: required non-null, non-empty, language-independent string. Its value is a service-defined error code. This code serves as a sub-status for the HTTP error code specified in the response.

In actuality, I recall seeing it either represent the actual HTTP error code in some cases, in other cases it is a workload provided code, and in other cases, our client sets it as Unknown. I should change the comment to be what it actual represents, not what it is supposed to. :(

peombwa: @MIchaelMainer LGTM! What was the conclusion on code property for Error object? /cc @darrelmiller

src/Microsoft.Graph.Core/Exceptions/ServiceException.cs
```diff
@@ -25,15 +35,23 @@ public ServiceException(Error error, System.Net.Http.Headers.HttpResponseHeaders
...
         /// <summary>
         /// Creates a new service exception.
         /// </summary>
         /// <param name="error">The error that triggered the exception.</param>
         /// <param name="innerException">The possible innerException.</param>
-        public ServiceException(Error error, Exception innerException = null)
+        /// <param name="responseHeaders">The HTTP response headers from the response.</param>
+        /// <param name="statusCode">The HTTP status code from the response.</param>
+        /// <param name="rawResponseBody">The raw JSON response body.</param>
+        public ServiceException(Error error,
```
andrueastman: It looks like calling this constructor will cause some loss of information. 
The `statusCode` and `responseHeaders` parameters are reset to `default` and `null` respectively and will therefore not be passed to the new object being created.

src/Microsoft.Graph.Core/Exceptions/Error.cs
darrelmiller: @MIchaelMainer   Hmm, weird, I thought I replied to this.  The top level code should never be a workload specific code moving forward.

darrelmiller: We need to change this comment.  We need to clarify that the value of the code depends on whether it is a top level Error or an innerError.

src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
```diff
@@ -19,11 +19,10 @@
     <DelaySign>false</DelaySign>
     <AssemblyOriginatorKeyFile>35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
-    <VersionPrefix>1.17.0</VersionPrefix>
+    <VersionPrefix>1.18.0</VersionPrefix>
     <VersionSuffix></VersionSuffix>
     <PackageReleaseNotes>
-      - Removed PageIterator. It is now available from the service library.
-      - Add support for time based retry limits.
+      - Improved ServiceException/error information to include rull OData error response, raw response body, and client-request-id.
```
darrelmiller: Typo "rull"?

src/Microsoft.Graph.Core/Exceptions/ErrorDetails.cs
```diff
@@ -0,0 +1,85 @@
+﻿// ------------------------------------------------------------------------------
```
darrelmiller: File is called ErrorDetails but the class is ErrorDetail.  Can we rename the file match the class?

src/Microsoft.Graph.Core/Exceptions/ServiceException.cs
darrelmiller: Agreed, this seems wrong.

src/Microsoft.Graph.Core/Exceptions/ServiceException.cs
```diff
@@ -53,6 +71,11 @@ public ServiceException(Error error, Exception innerException = null)
         /// </summary>
         public System.Net.HttpStatusCode StatusCode { get; }
 
+        /// <summary>
+        /// Provide the raw JSON response body.
+        /// </summary>
+        public string RawResponseBody { get; }
```
darrelmiller: If we have Error, which should be a strongly typed version of the body, what is the reason for also keeping the rawResponseBody around too?

src/Microsoft.Graph.Core/Exceptions/Error.cs
MIchaelMainer: Updated.

src/Microsoft.Graph.Core/Exceptions/ServiceException.cs
MIchaelMainer: A customer requested this, and we added this to our todo list. This was from this feature request: https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/316.

Fair enough question since it is all available in the Error object.

darrelmiller: I think the main point from that issue is that there were scenarios where we were not returning an error object at all.  e.g. An action that didn't have a return type.  I think returning a raw response was just the workaround solution

MIchaelMainer: We have a scenario where we swallow the error response returned in the response body when there is a ErrorResponse deserialization exception. We should provide the response body in that scenario.

src/Microsoft.Graph.Core/Exceptions/ServiceException.cs
MIchaelMainer: addressed

src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
MIchaelMainer: addressed

src/Microsoft.Graph.Core/Exceptions/ErrorDetails.cs
MIchaelMainer: addressed

Issue #27, opened 2019-04-26
HttpProvider re-write to fully support GraphClientFactory.


peombwa:
This will involve the completion of the following

- [x] Add a new provider to enable use of custom HttpClient and in turn maintain backward compatibility
- [x] Modify the SDK code generator to handle the new constructor

darrelmiller:
Enable support for this scenario:

```
  HttpClient client = GraphClientFactory.Create(...);
  var graphServiceClient = new GraphServiceClient(client);

```
Using a new provider something like this
```
  public class SimpleHttpProvider : IHttpProvider
    {
        private readonly HttpClient httpClient;

        public SimpleHttpProvider(HttpClient httpClient, ISerializer serializer = null)
        {
            this.httpClient = httpClient;
            Serializer = serializer ?? new Serializer();
            OverallTimeout = new TimeSpan(0, 5, 0);
        }
        public ISerializer Serializer { get; private set; }

        public TimeSpan OverallTimeout { get; set; }

        public void Dispose()
        {
            this.httpClient.Dispose();
        }

        public Task<HttpResponseMessage> SendAsync(HttpRequestMessage request)
        {
            return httpClient.SendAsync(request);
        }

        public Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, HttpCompletionOption completionOption, CancellationToken cancellationToken)
        {
            return httpClient.SendAsync(request, completionOption, cancellationToken);
        }
    }
```

PR #38, opened 2019-09-25
Add SimpleHttpProvider for GraphClientFactory compatibility

andrueastman:
- Adds a new IHttpProvider to enhance compatibility with GraphClientFactory by working towards adding support for the following scenario.
```
  HttpClient client = GraphClientFactory.Create(...);
  var graphServiceClient = new GraphServiceClient(client);

```
This means that one can easily use a custom httpClient with the request builders provided.
- Modifies checks in BaseRequests.cs to not authenticate the requests when the Provider is 
 either SimpleHttprovider or HttpProvider  and when authprovider is present.
- Adds tests

### Other links
- resolves a task in #27  
- is the predecessor for https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/221

### DIFF
--- a/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
+++ b/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
@@ -19,7 +19,7 @@
     <DelaySign>false</DelaySign>
     <AssemblyOriginatorKeyFile>35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
-    <VersionPrefix>1.17.0</VersionPrefix>
+    <VersionPrefix>1.18.0</VersionPrefix>
     <VersionSuffix></VersionSuffix>
     <PackageReleaseNotes>
       - Removed PageIterator. It is now available from the service library.
--- a/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
+++ b/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
@@ -19,11 +19,10 @@
     <DelaySign>false</DelaySign>
     <AssemblyOriginatorKeyFile>35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
-    <VersionPrefix>1.18.0</VersionPrefix>
+    <VersionPrefix>1.19.0</VersionPrefix>
     <VersionSuffix></VersionSuffix>
     <PackageReleaseNotes>
-      - Removed PageIterator. It is now available from the service library.
-      - Add support for time based retry limits.
+      - Resolves #27. Enhances GraphClientFactory compatibility by enabling GraphServiceClient creation by passing in a custom HttpClient
     </PackageReleaseNotes>
   </PropertyGroup>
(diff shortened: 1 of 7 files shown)


src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
```diff
@@ -19,7 +19,7 @@
     <DelaySign>false</DelaySign>
     <AssemblyOriginatorKeyFile>35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
-    <VersionPrefix>1.17.0</VersionPrefix>
+    <VersionPrefix>1.18.0</VersionPrefix>
     <VersionSuffix></VersionSuffix>
     <PackageReleaseNotes>
       - Removed PageIterator. It is now available from the service library.
```
MIchaelMainer: Remove old release notes. Add a release note for this work. Will be visible from NuGet.org and the package.

src/Microsoft.Graph.Core/Requests/SimpleHttpProvider.cs
```diff
@@ -0,0 +1,200 @@
...
+                        if (response.Headers.TryGetValues(CoreConstants.Headers.ThrowSiteHeaderName, out var throwSiteValues))
+                        {
+                            error.ThrowSite = throwSiteValues.FirstOrDefault();
+                        }
+                    }
+
+                    // Pass through the response headers and status code to the ServiceException.
+                    // System.Net.HttpStatusCode does not support RFC 6585, Additional HTTP Status Codes.
+                    // Throttling status code 429 is in RFC 6586. The status code 429 will be passed through.
+                    throw new ServiceException(error, response.Headers, response.StatusCode);
```
MIchaelMainer: Please review #37 as it adds some additional functionality to the HttpProvider that this provider should use to better support error/diagnostic info. Actually, I need a reviewer :). Let's pull 37 in and then we can pull yours in.

src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
andrueastman: Resolved

src/Microsoft.Graph.Core/Requests/SimpleHttpProvider.cs
andrueastman: That sounds like a plan.

andrueastman:
We will pull this in once #37 is completed and I have pull the necessary error/diagnostic changes

andrueastman:
New support for error/diagnostic info has now been incorporated. This PR should now be ready to be pulled in.

@MIchaelMainer @darrelmiller @irvinesunday

MIchaelMainer:
It looks fine - just set the version back to 1.18.0 and then we should be set. Part of me wants to turn some of those thrown ServiceException into ClientException as they aren't really ServiceException. Perhaps we do that wholesale at a later break.

andrueastman:
Version number has now been corrected to 1.18.0

darrelmiller:
Looks like merging the two branches broke some tests.  Seems related to ThrowSite. /cc @andrueastman  @MIchaelMainer

andrueastman:
Thanks pointing this out. This has been resolved. 

To prevent this from occurring again, I think we might need to look into the pr validation checks as it looks to me like the tests are not really run.

PR #44, opened 2019-10-09
Update master with latest preview features

MIchaelMainer:
Kicks off a production build. If all goes well, expect a deployment approval request for 1.18.0 after this PR is merged.


src/Microsoft.Graph.Core/Requests/BaseClient.cs
```diff
@@ -29,6 +30,19 @@ public class BaseClient : IBaseClient
             this.HttpProvider = httpProvider ?? new HttpProvider(new Serializer());
         }
 
+        /// <summary>
+        /// Constructs a new <see cref="BaseClient"/>.
+        /// </summary>
+        /// <param name="baseUrl">The base service URL. For example, "https://graph.microsoft.com/v1.0."</param>
+        /// <param name="httpClient">The custom <see cref="HttpClient"/> to be used for making requests</param>
+        public BaseClient(
+            string baseUrl,
```
darrelmiller: @MIchaelMainer @peombwa Should we remove the BaseUrl parameter as HttpClient has a BaseAddress property?  Or is it better to keep it there for consistency?

MIchaelMainer: I think we need to keep this. We don't use the BaseAddress set on HttpClient in the service library. We build the URL from the BaseClient.baseUrl IIRC.

We would need to change the PR in the generator if we remove the BaseUrl parameter.

peombwa: @darrelmiller I agree with Michael, let's keep it for now. Everything else LGTM.

PR #47, opened 2019-10-11
BatchRequestStep enhancements

### LINKED ISSUE #45: BatchRequestStep enhancements
a) Add a method to BatchRequestContent that is an overload of AddBatchRequestStep that accepts HttpRequestMessage as a parameter without having to provide an Id.  Generate an Id using Guid.NewGuid.

b) Add another overload that accepts anything derived from baseRequest

The goal is to simplify creating batches for the simple case.  Developers should be able to use GraphServiceClient to create request objects and just add them to a BatchRequestContent instance.

/cc @MIchaelMainer

andrueastman:
This PR closes #45

1. Adds overload to add steps using HttpRequestMessage to support a scenario like
```cs
HttpRequestMessage httpRequestMessage1 = new HttpRequestMessage(HttpMethod.Get, "https://graph.microsoft.com/v1.0/me/");
HttpRequestMessage httpRequestMessage2 = new HttpRequestMessage(HttpMethod.Get, "https://graph.microsoft.com/v1.0/me/calendars");

// Add batch request steps to BatchRequestContent.
BatchRequestContent batchRequestContent = new BatchRequestContent();
batchRequestContent.AddBatchRequestStep(httpRequestMessage1);
batchRequestContent.AddBatchRequestStep(httpRequestMessage2);

// Send batch request with BatchRequestContent. 
HttpResponseMessage response = await httpClient.PostAsync("https://graph.microsoft.com/v1.0/$batch", batchRequestContent);
```
2. Adds overload to add steps using IBaseRequest to support a scenario like
```cs
IUserRequest request1 = graphServiceClient.Me.Request();
IUserCalendarsCollectionRequest request2 = graphServiceClient.Me.Calendars.Request();

// Add batch request steps to BatchRequestContent.
BatchRequestContent batchRequestContent = new BatchRequestContent();
batchRequestContent.AddBatchRequestStep(request1);
batchRequestContent.AddBatchRequestStep(request2);

// Send batch request with BatchRequestContent. 
HttpResponseMessage response = await httpClient.PostAsync("https://graph.microsoft.com/v1.0/$batch", batchRequestContent);
```
3. Adds tests to validate the behavior and to ensure BatchRequestSteps do not exceed max value set for the batch endpoint

### DIFF
--- a/src/Microsoft.Graph.Core/Requests/Content/BatchRequestContent.cs
+++ b/src/Microsoft.Graph.Core/Requests/Content/BatchRequestContent.cs
@@ -1,4 +1,4 @@
-﻿// ------------------------------------------------------------------------------
+// ------------------------------------------------------------------------------
 //  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
 // ------------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Graph.Core/Requests/Content/BatchRequestContent.cs
```diff
@@ -78,8 +78,46 @@ public BatchRequestContent(params BatchRequestStep[] batchRequestSteps)
         /// <returns>True or false based on addition or not addition of the provided <see cref="BatchRequestStep"/>. </returns>
         public bool AddBatchRequestStep(BatchRequestStep batchRequestStep)
```
darrelmiller: When we do a breaking change we should consider changing this to return void and throw an exception rather than return bool.  Requiring a consumer to check a return value is a `C` style thing to do and not something a C# dev would expect.

src/Microsoft.Graph.Core/Requests/Content/BatchRequestContent.cs
```diff
@@ -78,8 +78,46 @@ public BatchRequestContent(params BatchRequestStep[] batchRequestSteps)
...
+            (BatchRequestSteps as IDictionary<string, BatchRequestStep>).Add(batchRequestStep.RequestId, batchRequestStep);
+            return true;
+        }
+
+        /// <summary>
+        /// Adds a <see cref="HttpRequestMessage"/> to batch request content.
+        /// </summary>
+        /// <param name="httpRequestMessage">A <see cref="HttpRequestMessage"/> to use to build a <see cref="BatchRequestStep"/> to add.</param>
+        /// <returns>True or false based on addition or not addition of the provided <see cref="HttpRequestMessage"/>. </returns>
+        public bool AddBatchRequestStep(HttpRequestMessage httpRequestMessage)
```
darrelmiller: I'm tempted to suggest we return void here instead of bool.  Returning bool significantly complicates the calling code.

peombwa: Since we are moving away from returning bool, I see no issue with returning void and throwing exceptions if MaxNumberOfRequests are exceeded for the new overloads. The intended caller of the new overloads would probably be having a try catch block for graphServiceClient calls. Consistency will be lost, but breaking changes for our customers will be limited to the `AddBatchRequestStep(BatchRequestStep batchRequestStep)` overload.
/cc @darrelmiller, @MIchaelMainer, @andrueastman

peombwa: Also, BatchRequestContent ctor already throws a ClientException with MaximumValueExceeded code.

andrueastman: This makes sense to me. 
I will resolve this by returning void in the new overloads, and leave the changing of the existing `AddBatchRequestStep(BatchRequestStep batchRequestStep)` overload be be sorted out in the next time we do a breaking change.

andrueastman: This should be resolved now.

src/Microsoft.Graph.Core/Requests/Content/BatchRequestContent.cs
andrueastman: New issue created to track this https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/issues/49

andrueastman:
@MIchaelMainer @peombwa @darrelmiller This should be ready to review. 
https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/issues/49 will track updating the already existing `AddBatchRequestStep(BatchRequestStep batchRequestStep)` overload on the next breaking change.

PR #58, opened 2019-10-29
DeltaResponseHandler adds a delta response payload manifest to CollectionPage.AdditionalData

MIchaelMainer:
[Delta query](https://docs.microsoft.com/en-us/graph/delta-query-users?view=graph-rest-1.0) provides a way to track changes.  You begin to track changes with a request like this:

`GET https://graph.microsoft.com/v1.0/users/delta?$select=displayName,givenName,surname`

`var userDeltaCollectionPage = await graphClient.Users.Delta().Request().Select("displayName,givenName,surname").GetAsync();`

Currently, a developer has to inspect the selected properties on each object in the page. The inspection requires first that they check that the selected properties are not null before getting the changed value. If the selected property is a primitive, then they can capture that value. If the selected property is an object, they must also inspect all of its property with a null check before capturing the values.  This may mean unnecessarily checking many properties that have not changed. This PR should reduce the amount of null checking that has to occur when a delta query payload is received by providing a manifest of changes. This is especially useful when a property has been set to null as previously, there was no way to determine whether a property has been set to null, or was not set and so it defaulted to null. 

The change manifest is set on each item in the CollectionPage. It is found in the AdditionalData dictionary with a a key called `changes`.  This functionality is accessed via:

**Microsoft.Graph.Core**

```csharp
var deltaServiceLibResponse = await deltaResponseHandler.HandleResponse<JObject>(httpresponsemessage);
var changesOnTheFirstItem = deltaJObjectResponse["value"][0]["changes"] as JArray;

// This is not an important scenario as the developer can query whether a value is set null using JMESPath or using Newtonsoft.
```

**Microsoft.Graph**

This enables developers to discover values set to null when using the service library. 

```csharp
var eventsDeltaCollectionPage = await graphClient.Me.CalendarView
                                               .Delta()
                                               .Request()
                                               .WithResponseHandler(new DeltaResponseHandler())
                                               .Select("subject")
                                               .GetAsync();

// Accesses the changes on each page item.
eventsDeltaCollectionPage[0].AdditionalData.TryGetValue("changes", out object changes);
var changeList = (changes as JArray).ToObject<List<string>>();

// Updating a non-schematized property on a model such as instance annotations, open types,
// and schema extensions. Can also be used to discover deleted items.
if (changeList.Exists(x => x.Equals("@removed.reason")))
{
    eventsDeltaCollectionPage[0].AdditionalData.TryGetValue("@removed", out object odataEtag);
    Console.Writeline("This item was removed.")
    // TODO: myModel.Delete();
}

// Core scenario - update schematized property regardless of it is set to null.
// This property has been set to null in the response. We can be confident that
// whatever the value set is correct, regardless whether it is null.
if (changeList.Exists(x => x.Equals("subject")))
{
    myModel.Subject = eventsDeltaCollectionPage[0].Subject;
}

// Update the value on a complex type property's value. Developer can't just replace the body
// as that could result in overwriting other unchanged property. Essentially, they need to inspect
// every leaf node in the selected property set.
if (changeList.Exists(x => x.Equals("body.content"))) // 
{
    if (myModel.Body == null)
    {
        myModel.Body = new ItemBody();
    }

    myModel.Body.Content = eventsDeltaCollectionPage[0].Body.Content;
}

// Update complex type property's value when the value is a collection of objects.
// We don't know whether this is an update or add without querying the client model.
// We will need to check each object in the model.
var attendeesChangelist = changeList.FindAll(x => x.Contains("attendees"));
if (attendeesChangelist.Count > 0)
{
    // This is where if we provided the delta response as a JSON object, 
    // we could let the developer use JMESPath to query the changes.
    if (changeList.Exists(x => x.Equals("attendees[0].emailAddress.name")))
    {
        if (myModel.Attendees == null) // Attendees are being added for the first time.
        {
            var attendees = new List<Attendee>();
            attendees.AddRange(eventsDeltaCollectionPage[0].Attendees);
            myModel.Attendees = attendees;
        }
        else // Attendees list is being updated.
        {
            // We need to inspect each object, and determine which objects and properties 
            // need to be initialized and/or updated.
        }
    }
}

```

The raw JSON on the object after adding the change list looks like:

```json
{
    "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#Collection(event)",
    "@odata.nextLink": "https://graph.microsoft.com/v1.0/me/calendarView/delta?$skiptoken=R0usmci39OQxqJrxK4",
    "value": [
        {
            "@odata.type": "#microsoft.graph.event",
            "@odata.etag": "EZ9r3czxY0m2jz8c45czkwAAFXcvIw==",
            "subject": null,
            "id": "AAMkADVxTAAA=",
            "changes" : ["@odata.type", "@odata.etag", "subject", "id"]
        }
    ]
}
 ```

This also now allows developers to add their own response handler to any request builder. This enables new ways to process the contents of a request builder based response. 

Generally, a change results in the returning the entire item with the original property set. 

> The delta feed shows the latest state for each item, not each change. If an item were renamed twice, it would only show up once, with its latest name.

https://docs.microsoft.com/en-us/graph/api/driveitem-delta?view=graph-rest-1.0&tabs=http

- [x] Put changes on the object, not the page collection. Will need to JObject, not JsonTextReader
- [x] Should work with only core and also with service libraries.
- [x] support open properties in the change list.

### DIFF
--- a/src/Microsoft.Graph.Core/Requests/DeltaResponseHandler.cs
+++ b/src/Microsoft.Graph.Core/Requests/DeltaResponseHandler.cs
@@ -4,9 +4,9 @@
 
 namespace Microsoft.Graph
 {
-    using Newtonsoft.Json;
+    using Newtonsoft.Json.Linq;
+    using System;
     using System.Collections.Generic;
-    using System.IO;
     using System.Linq;
     using System.Net.Http;
     using System.Threading.Tasks;
@@ -21,11 +21,10 @@ namespace Microsoft.Graph
     public class DeltaResponseHandler : IResponseHandler
     {
         private readonly ISerializer serializer;
+
         /// <summary>
         /// Constructs a new <see cref="ResponseHandler"/>.
         /// </summary>
-        /// <param name="serializer"></param>
-        
         public DeltaResponseHandler()
         {
             this.serializer = new Serializer();
@@ -37,46 +36,22 @@ namespace Microsoft.Graph
         /// <typeparam name="T">The type to return</typeparam>
         /// <param name="response">The HttpResponseMessage to handle</param>
         /// <returns></returns>
-        public async Task<T> HandleResponse<T>(HttpResponseMessage response) 
+        public async Task<T> HandleResponse<T>(HttpResponseMessage response)
         {
             if (response.Content != null)
             {
+                // Gets the response string with response headers and status code
+                // set on the response body object.
                 var responseString = await GetResponseString(response);
 
-                return serializer.DeserializeObject<T>(responseString);
+                // Get the response body object with the change list 
+                // set on each response item.
+                JObject responseBody = await GetResponseBodyWithChangelist(responseString);
+
+                return responseBody.ToObject<T>();
             }
 
             return default(T);
-        }
-
-        /// <summary>
-        /// Get the change properties.
-        /// </summary>
-        /// <param name="responseString">The response from the service.</param>
-        /// <returns>A list of changes.</returns>
-        private List<string> GetChangedProperties(string responseString)
-        {
-            var changedPropertiesList = new List<string>();
-
-            using (JsonTextReader reader = new JsonTextReader(new StringReader(responseString)))
-            {
-                // Forward the reader to the list of changes.
-                while (reader.TokenType != JsonToken.StartArray) // This is a weak assumption. 
-                {
-                    reader.Read();
-                }
-                
-                // Read all of the changed properties in the page and
-                // add them to the changed property list.
-                while (reader.Read())
-                {
-                    if (reader.TokenType == JsonToken.PropertyName)
-                    {
-                        changedPropertiesList.Add(reader.Path);
-                    }
-                }
-            }
-            return changedPropertiesList;
         }
 
         /// <summary>
@@ -93,10 +68,6 @@ namespace Microsoft.Graph
             //Only add headers and the changed list if we are going to return a response body
             if (content.Length > 0)
             {
-                // Get the list of changes in the delta response. 
-                List<string> changedProperties = GetChangedProperties(content);
-                var changes = serializer.SerializeObject(changedProperties);
-
                 // Add headers
                 var responseHeaders = hrm.Headers;
                 var statusCode = hrm.StatusCode;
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/Microsoft.Graph.Core/Requests/DeltaResponseHandler.cs [hunk 1]
```diff
@@ -0,0 +1,116 @@
...
+    /// deserializer can't express changes to null so you can now discover if a property
+    /// has been set to null. This is intended for use with a Delta query scenario.
+    /// </summary>
+    public class DeltaResponseHandler : IResponseHandler
+    {
+        private readonly ISerializer serializer;
+        /// <summary>
+        /// Constructs a new <see cref="ResponseHandler"/>.
+        /// </summary>
+        /// <param name="serializer"></param>
```
andrueastman: param `serializer`  declared in comment does not exist.

tests/Microsoft.Graph.DotnetCore.Core.Test/Requests/ResponseHandlerTests.cs
```diff
@@ -38,5 +41,80 @@ public async Task HandleUserResponse()
...
+
+            // Act
+            var deltaResponse = await deltaResponseHandler.HandleResponse<EventDeltaCollectionResponse>(hrm);
+            bool hasChanges = deltaResponse.AdditionalData.TryGetValue("changes", out object changes);
+
+            var serializer = new Serializer();
+            List<string> changeList = serializer.DeserializeObject<List<string>>(changes.ToString());
+
+            // Assert
+            Assert.True(hasChanges);
```
andrueastman: If the "changeList" is indeed empty, should we have changes? Or should `hasChanges` be false?

andrueastman:
As suggested, I agree it would be a better experience to provide a way to return an instance of ` List<List<string>>` with the first level being the list of change objects and the second level being the list of strings of changed properties in dot notation.

I think this may also make it easier for the user to correlate the `changes` property list to the actual values in the DeltaCollectionPage. The user could possibly use the index in the 'changes' list to retrieve the corresponding actual value using the same index in the DeltaCollectionPage.

src/Microsoft.Graph.Core/Requests/DeltaResponseHandler.cs
```diff
@@ -0,0 +1,116 @@
...
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// PREVIEW 
+    /// A response handler that exposes the list of changes returned in a response.
+    /// This supports scenarios where the service expresses changes to 'null'. The
+    /// deserializer can't express changes to null so you can now discover if a property
+    /// has been set to null. This is intended for use with a Delta query scenario.
+    /// </summary>
+    public class DeltaResponseHandler : IResponseHandler
```
MIchaelMainer: Add docs on how to use this. Ideally, this is only used after an initial sync, as we have to inspect the contents of the response twice to determine the change list.

MIchaelMainer:
@andtu for future review

tests/Microsoft.Graph.DotnetCore.Core.Test/Requests/ResponseHandlerTests.cs
```diff
@@ -38,5 +42,188 @@ public async Task HandleUserResponse()
             var headers = (JObject)(user.AdditionalData["responseHeaders"]);
             Assert.Equal("value", (string)headers["test"][0]);
         }
+
```
andrueastman: I think it would be great if we could have a scenario that has ```@removed": {"reason": "changed"}```
in the payload to validate/illustrate when an object is in the delta due to deletion.

peombwa: I agree with this, @MIchaelMainer.

peombwa: Otherwise, it LGTM.

tests/Microsoft.Graph.DotnetCore.Core.Test/Requests/ResponseHandlerTests.cs
```diff
@@ -38,5 +42,188 @@ public async Task HandleUserResponse()
...
+            // IEventDeltaCollectionPage is what the service library provides.
+            // Can't test against the service library model, since it has a reference to the signed
+            // public version of this library. see issue #57 for more info.
+            // https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/issues/57
+            // Service library testing will need to happen in the service library repo once this is published on NuGet.
+
+            // Assert
+            Assert.True(deltaServiceLibResponse.Value is IEventDeltaCollectionPage); // We create a valid ICollectionPage.
+            Assert.Equal("George", attendeeName); // We maintain the expected response body when we change it.
+            Assert.Equal("attendees[0].emailaddress.name", attendeeNameInChangelist); // We expect that this property is in changelist.
```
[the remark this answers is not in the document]
MIchaelMainer: Addressed in 3d2e1d12

tests/Microsoft.Graph.DotnetCore.Core.Test/Requests/ResponseHandlerTests.cs
MIchaelMainer: Addressed in 3d2e1d12

src/Microsoft.Graph.Core/Requests/IBaseRequest.cs
```diff
@@ -12,6 +12,11 @@ namespace Microsoft.Graph
     /// </summary>
     public interface IBaseRequest
     {
+        /// <summary>
+        /// Gets or sets the response handler for the request.
+        /// </summary>
+        IResponseHandler ResponseHandler { get; set; }
```
darrelmiller: This is going to be a breaking change for anyone who has implemented their own Request object that implements IBaseRequest.

MIchaelMainer: Thank you for catching this. I'm confident that this will break someone out there.

MIchaelMainer: Fixed, we just needed this on BaseRequest.

PR #73, opened 2019-12-23
Fix default upload size (5 instead of 4)

marcoscheel:
If LargeFileUploadTask is called without the maxSliceSize parameter the default will be picked. But the default was 4*1024*1024 and that is not a multiplier of 320! So a caller must specify the size, even if they want to stick to the defaults (that may be changed in the future).

Fixes #

-
-
-

### Other links
-
-
-


andrueastman:
@marcoscheel
Thanks for taking the time to raise this issue.
I have captured the change in #159 so that the build can pass. I will close this PR for now.
Thanks again for raising this.

PR #95, opened 2020-03-03
TokenCredential Support

andrueastman:
1. Adds support for TokenCredentials from Azure.Identity by adding TokenCredential Auth provider.

Therefore one can make use a TokenCredential from Azure.Identity to make calls like the example below which uses the InteractiveBrowserCredential.

```cs
string clientId = "clientId";
string[] scopes = new[] { "User.Read", "Mail.ReadWrite" };

InteractiveBrowserCredential myBrowserCredential = new InteractiveBrowserCredential(clientId);
TokenCredentialAuthProvider tokenCredentialAuthProvider = new TokenCredentialAuthProvider(myBrowserCredential, scopes);

//Try to get something from the Graph!!
HttpClient httpClient = GraphClientFactory.Create(tokenCredentialAuthProvider);
HttpRequestMessage requestMessage = new HttpRequestMessage(HttpMethod.Get, "https://graph.microsoft.com/v1.0/me/");
HttpResponseMessage response = await httpClient.SendAsync(requestMessage);
```

2. Adds baseclient constructor to take in tokencredential to the baseclient
This is done so as to enable a future case in the service library that would looks as below
```cs
string[] scopes = new[] { "User.Read", "Mail.ReadWrite" };
InteractiveBrowserCredential myBrowserCredential = new InteractiveBrowserCredential(clientId);

//Try to get something from the Graph!!
GraphServiceClient graphClient = new GraphServiceClient(myBrowserCredential);

var user = await graphClient.Me.Request().WithScopes(scopes).GetAsync();
```
3. Updates runtimes to .net standard 2.0
4. Adds an IntergratedWindowsTokenCredential not provided by Azure.Identity

This can be used with the core library as follows
```cs
string clientId = "clientId";
string authority = "https://login.microsoftonline.com/organizations/";
string[] scopes = new[] { "User.Read", "Mail.ReadWrite" };

IPublicClientApplication publicClientApplication = PublicClientApplicationBuilder
    .Create(clientId)
    .Build();

IntegratedWindowsTokenCredential integratedWindowsTokenCredential = new IntegratedWindowsTokenCredential(publicClientApplication);
TokenCredentialAuthProvider tokenCredentialAuthProvider = new TokenCredentialAuthProvider(integratedWindowsTokenCredential, scopes);

//Try to get something from the Graph!!
HttpClient httpClient = GraphClientFactory.Create(tokenCredentialAuthProvider);

HttpRequestMessage requestMessage = new HttpRequestMessage(HttpMethod.Get, "https://graph.microsoft.com/v1.0/me");
```
5. Tests have been added to validate the various changes made above.

### DIFF
--- a/src/Microsoft.Graph.Core/Authentication/TokenCredentials/IntegratedWindowsTokenCredential.cs
+++ b/src/Microsoft.Graph.Core/Authentication/TokenCredentials/IntegratedWindowsTokenCredential.cs
@@ -30,14 +30,12 @@ namespace Microsoft.Graph
         /// Creates a new IntegratedWindowsTokenCredential which will authenticate users with the specified application.
         /// </summary>
         /// <param name="publicClientApplication">A <see cref="IPublicClientApplication"/> to pass to <see cref="IntegratedWindowsTokenCredential"/> for authentication.</param>
+        /// <exception cref="ArgumentException"> When a null <see cref="IPublicClientApplication"/> is passed</exception>
         public IntegratedWindowsTokenCredential(IPublicClientApplication publicClientApplication)
         {
-            _clientApplication = publicClientApplication ?? throw new AuthenticationException(
-                new Error
-                {
-                    Code = ErrorConstants.Codes.InvalidRequest,
-                    Message = string.Format(ErrorConstants.Messages.NullValue, nameof(publicClientApplication))
-                });
+            _clientApplication = publicClientApplication ?? throw new ArgumentException(
+                                     string.Format(ErrorConstants.Messages.NullParameter, nameof(publicClientApplication)),
+                                     nameof(publicClientApplication));
         }
 
         /// <summary>
@@ -57,6 +55,7 @@ namespace Microsoft.Graph
         /// <param name="requestContext">The <see cref="TokenRequestContext"/> for the request</param>
         /// <param name="cancellationToken">The <see cref="CancellationToken"/> for the request</param>
         /// <returns>An <see cref="AccessToken"/> to make requests with</returns>
+        /// <exception cref="AuthenticationException"> When an unknown authentication exception occurs</exception>
         public override async ValueTask<AccessToken> GetTokenAsync(TokenRequestContext requestContext, CancellationToken cancellationToken)
         {
             AuthenticationResult result;
@@ -69,9 +68,6 @@ namespace Microsoft.Graph
                 // Try to get the token silently
                 AcquireTokenSilentParameterBuilder tokenSilentBuilder =
                     _clientApplication.AcquireTokenSilent(requestContext.Scopes, account);
-
-                if (!ContainsWellKnownTenantName(_clientApplication.Authority))
-                    tokenSilentBuilder.WithAuthority(_clientApplication.Authority);
 
                 result = await tokenSilentBuilder.ExecuteAsync(cancellationToken);
             }
@@ -99,6 +95,7 @@ namespace Microsoft.Graph
                 {
                     authenticationResult = await _clientApplication.AcquireTokenByIntegratedWindowsAuth(requestContext.Scopes)
                                 .ExecuteAsync();
+                    break;
                 }
                 catch (MsalServiceException serviceException)
                 {
@@ -137,19 +134,5 @@ namespace Microsoft.Graph
             return authenticationResult;
         }
 
-        /// <summary>
-        /// Determines if an authority url has a `well known` tenant name (common, consumers, organizations) in its tenant segment.
-        /// </summary>
-        /// <param name="currentAuthority">The authority url to check.</param>
-        /// <returns>True is pre-defined tenant names are present, and false if not. </returns>
-        private bool ContainsWellKnownTenantName(string currentAuthority)
-        {
-            if (!Uri.IsWellFormedUriString(currentAuthority, UriKind.Absolute))
-                throw new ArgumentException("Invalid authority URI set.");
-
-            Uri authorityUri = new Uri(currentAuthority);
-
-            return WellKnownTenants.Contains(authorityUri.Segments[1].Replace(@"/", string.Empty));
-        }
     }
 }
--- a/src/Microsoft.Graph.Core/Authentication/TokenCredentials/IntegratedWindowsTokenCredential.cs
+++ b/src/Microsoft.Graph.Core/Authentication/TokenCredentials/IntegratedWindowsTokenCredential.cs
@@ -1,0 +1,155 @@
+﻿// ------------------------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+// ------------------------------------------------------------------------------
+
+namespace Microsoft.Graph
+{
+    using Azure.Core;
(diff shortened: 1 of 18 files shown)


src/Microsoft.Graph.Core/Authentication/TokenCredentials/IntegratedWindowsTokenCredential.cs
```diff
@@ -0,0 +1,155 @@
...
+            {
+                // Try to get the account from the cache
+                IEnumerable<IAccount> accounts = await _clientApplication.GetAccountsAsync();
+                IAccount account = accounts.FirstOrDefault();
+
+                // Try to get the token silently
+                AcquireTokenSilentParameterBuilder tokenSilentBuilder =
+                    _clientApplication.AcquireTokenSilent(requestContext.Scopes, account);
+
+                if (!ContainsWellKnownTenantName(_clientApplication.Authority))
```
MIchaelMainer: I don't understand this check. 

If the authority is one of the well known endpoints, the _clientApplication.Authority is used (I assume it is used by `tokenSilentBuilder`). If it isn't, as in `login.microsoftonline.com/{tenantId}`, then the same _clientApplication.Authority is set on the `tokenSilentBuilder`. What is the difference of setting the authority on the `_clientApplication` versus on `tokenSilentBuilder`?

ContainsWellKnownTenantName seems to be more of a check on the wellformedness of the authority URI.

src/Microsoft.Graph.Core/Authentication/TokenCredentials/IntegratedWindowsTokenCredential.cs
```diff
@@ -0,0 +1,155 @@
...
+        private async Task<AuthenticationResult> GetNewAccessTokenAsync(TokenRequestContext requestContext)
+        {
+            AuthenticationResult authenticationResult = null;
+            int retryCount = 0;
+            do
+            {
+                try
+                {
+                    authenticationResult = await _clientApplication.AcquireTokenByIntegratedWindowsAuth(requestContext.Scopes)
+                                .ExecuteAsync();
```
MIchaelMainer: If we get an authenticationResult, then we should break from the do..while.

src/Microsoft.Graph.Core/Authentication/TokenCredentials/IntegratedWindowsTokenCredential.cs
```diff
@@ -0,0 +1,155 @@
...
+
+        private readonly IPublicClientApplication _clientApplication;
+
+        /// <summary>
+        /// Creates a new IntegratedWindowsTokenCredential which will authenticate users with the specified application.
+        /// </summary>
+        /// <param name="publicClientApplication">A <see cref="IPublicClientApplication"/> to pass to <see cref="IntegratedWindowsTokenCredential"/> for authentication.</param>
+        public IntegratedWindowsTokenCredential(IPublicClientApplication publicClientApplication)
+        {
+            _clientApplication = publicClientApplication ?? throw new AuthenticationException(
```
MIchaelMainer: We should throw a ArgumentNullException if publicClientApplication is null. We haven't attempted to authenticate yet at this point. Also, we should document `///<exception></exception>` thrown exceptions

src/Microsoft.Graph.Core/Authentication/TokenCredentials/IntegratedWindowsTokenCredential.cs
```diff
@@ -0,0 +1,155 @@
...
+                    if (serviceException.ErrorCode == ErrorConstants.Codes.TemporarilyUnavailable)
+                    {
+                        TimeSpan delay = this.GetRetryAfter(serviceException);
+                        retryCount++;
+                        // pause execution
+                        await Task.Delay(delay);
+                    }
+                    else
+                    {
+                        throw new AuthenticationException(
```
MIchaelMainer: Document this exception at the first public method where this can be encountered

src/Microsoft.Graph.Core/Exceptions/AuthenticationException.cs
```diff
@@ -0,0 +1,29 @@
+﻿// ------------------------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+// ------------------------------------------------------------------------------
+
+namespace Microsoft.Graph
+{
+    using System;
+    /// <summary>
+    /// Graph authentication exception
```
MIchaelMainer: This isn't really Graph authentication exception. This is an unknown authentication exception.. when MSAL throws an unknown exception, as in not an MSALException.

src/Microsoft.Graph.Core/Extensions/TokenCredentialExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+            {
+                delay = retryAfter.Delta;
+            }
+            else if (retryAfter != null && retryAfter.Date.HasValue)
+            {
+                delay = retryAfter.Date.Value.Offset;
+            }
+
+            if (delay == null)
+                throw new MsalServiceException(serviceException.ErrorCode, ErrorConstants.Messages.MissingRetryAfterHeader);
```
MIchaelMainer: Add the innerException that caused this.

src/Microsoft.Graph.Core/Authentication/TokenCredentials/IntegratedWindowsTokenCredential.cs
andrueastman: Well spotted! Thanks.

src/Microsoft.Graph.Core/Authentication/TokenCredentials/IntegratedWindowsTokenCredential.cs
andrueastman: You are correct. 
Since the user will set the `Authority` at the client level there is no need to do this. This should ideally be use if a well known Authority name is used at the top level and one wishes to narrow down the tenant name when fetching the token silently.

andrueastman: I have actioned on this and removed the check.

src/Microsoft.Graph.Core/Extensions/BaseRequestExtensions.cs
```diff
@@ -167,5 +167,41 @@ public static class BaseRequestExtensions
...
+        /// If you use a custom authentication handler or authentication provider, you have to handle it's retrieval in your implementation.
+        /// </summary>
+        /// <param name="baseRequest">The <see cref="IBaseRequest"/>.</param>
+        /// <param name="scopes">Microsoft graph scopes used to authenticate this request.</param>
+        public static T WithScopes<T>(this T baseRequest, string[] scopes) where T : IBaseRequest
+        {
+            string authHandlerOptionKey = typeof(AuthenticationHandlerOption).ToString();
+            AuthenticationHandlerOption authHandlerOptions; 
+
+            // make sure that the options exist in the middleware otherwise create is
```
ddyett: >is [](start = 83, length = 2)

NIT: it

PR #99, opened 2020-03-23
Support native plaform Http handlers for Xamarin.Mac

### LINKED ISSUE #413: MAUI support
Currently, this SDK doesn't work well on Android with MAUI, see details here (and a workaround): https://github.com/dotnet/maui/issues/3903

Is there any plans to add support for MAUI?

tipa:
This PR adds support for native http handler when targeting Xamarin.Mac (NSUrlSessionHandler)

### Fixes

#413 Cannot download multiple OneDrive files concurrently since v.1.13.0-beta

### DIFF
--- a/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
+++ b/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
@@ -4,7 +4,7 @@
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <AssemblyTitle>Microsoft Graph Core Client Library</AssemblyTitle>
     <Authors>Microsoft</Authors>
-    <TargetFrameworks>netstandard1.1;net45;Xamarin.iOS10;MonoAndroid70</TargetFrameworks>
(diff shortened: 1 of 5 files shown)


tipa:
Any timeline when the msgraph-sdk-dontnet package will use v1.21.0 which contains this change so I can stop using a workaround in my code?

MIchaelMainer:
@tipa Thank you for following up. Can you confirm that this works as expected in this preview: https://www.nuget.org/packages/Microsoft.Graph.Core/1.21.0-preview.1

tipa:
Unfortunately I cannot test it at the moment as VS for Mac doesn't start on my updated Big Sur machine (https://developercommunity.visualstudio.com/content/problem/1088817/visual-studio-2019-864-crashes-on-mac-os-big-sur.html).
Is it possible the change was merged into 1.20.1 already? That's the package that I am currently using and it seems to work as expected.

MIchaelMainer:
No, it isn't in 1.20.1. You have to be using your workaround for that.

PR #103, opened 2020-03-26
Release 1.20.1 - fix DeltaResponseHandler

MIchaelMainer:
Contains the fix for DeltaResponseHandler and CI/CD updates.

This will result in a release for 1.21.1.

@andrueastman I suggest you add a task to cherry pick c392714.

### DIFF
--- a/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
+++ b/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
@@ -23,6 +23,7 @@
     <VersionSuffix></VersionSuffix>
     <PackageReleaseNotes>
       - Fix for modifying the Timeout property of a passed HttpClient
+      - Add DeltaResponseHandler to discover null changes for delta query.
     </PackageReleaseNotes>
   </PropertyGroup>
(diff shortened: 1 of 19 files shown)


src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
```diff
@@ -19,12 +19,10 @@
     <DelaySign>false</DelaySign>
     <AssemblyOriginatorKeyFile>35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
-    <VersionPrefix>1.20.0</VersionPrefix>
+    <VersionPrefix>1.20.1</VersionPrefix>
```
ddyett: i think 1.20.1 is correct but the title is indicating this should be 1.21.1

MIchaelMainer: Fixed that. https://www.nuget.org/packages/Microsoft.Graph.Core 1.20.0 is the last release for this major version.

MIchaelMainer:
- [ ] Re-enable tests for all of these yaml pipelines.

MIchaelMainer:
One last change, I swear. Need to enable tests on building for preview and production build pipelines.

Issue #80, opened 2020-01-08
Support move to System.Text.Json on the Service Library 


andrueastman:
The service library currently uses the Newtonsoft.Json dependency from core. To have the service library move to System.Text.Json we need to achieve the following.

- [x] Update the code generator to emit the proper attributes for System.Text.Json in the generated models
- [x] Rewrite any model extensions [here](https://github.com/microsoftgraph/msgraph-sdk-dotnet/tree/dev/src/Microsoft.Graph/Models/Extensions) to support System.Text.Json
- [x] Rewrite the [ComplexTypeHelper.cs](https://github.com/microsoftgraph/msgraph-sdk-dotnet/blob/dev/src/Microsoft.Graph/Models/Helpers/ComplexTypeHelper.cs) to support System.Text.Json
- [x] Refactor any tests in Service Library with a Newtonsoft dependency 
[AB#6305](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6305)

PR #104, opened 2020-03-26
Add general support for System.Text.Json

andrueastman:
This PR adds general support for System.Text,Json to the Core Library moving away from NewtonSoft.

Only the DeltaResponseHandler is left that references Newtonsoft which will be addressed in an upcoming PR.

- Rewrite all converters to support system.Text.Json
- Update attributes in tests and System.Text.Json atrributes
- Create InterfaceConverter attribute to support adding attributes to interfaces(currently not supported in System.Text.Json)
- Refactored test models to enable validation of changes.

### Other links
- https://github.com/andrueastman/system-text-json-samples (Sample project using this PR as a submodule to show how changes in ServiceLibrary would look like and also help in validation)


src/Microsoft.Graph.Core/Serialization/DateTimeOffsetConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        /// <summary>
+        /// Writes the JSON representation of the object.
+        /// </summary>
+        /// <param name="writer">The <see cref="Utf8JsonWriter"/> to write to.</param>
+        /// <param name="dateTimeValue">The dateTime value.</param>
+        /// <param name="options">The calling serializer options</param>
+        public override void Write(Utf8JsonWriter writer, DateTimeOffset dateTimeValue, JsonSerializerOptions options)
+        {
+            writer.WriteStringValue(dateTimeValue.ToString());
```
ddyett: >dateTimeValue [](start = 36, length = 13)

null check?

ddyett:
> using System;

missing header for license.

---
Refers to: src/Microsoft.Graph.Core/Serialization/InterfaceConverterAttribute.cs:1 in b8a1d39. [](commit_id = b8a1d3939d0799ec8ce78f623ba5772784065ba5, deletion_comment = False)

tests/Microsoft.Graph.DotnetCore.Core.Test/Exceptions/ErrorTests.cs
```diff
@@ -126,11 +126,13 @@ public void Validate_ErrorObjectDeserializes()
             Assert.Equal("inner-error-code", error.InnerError.Code);
             Assert.Equal("target-value", error.Target);
             Assert.NotNull(error.AdditionalData);
-            Assert.Equal("unexpected-property-value", error.AdditionalData["unexpected-property"]);
+            var expectedValue1 = error.AdditionalData["unexpected-property"].ToString();
```
ddyett: >expectedValue1 [](start = 16, length = 14)

why introduce the variable instead of just adding the ToString() in the Assert.Equal

tests/Microsoft.Graph.DotnetCore.Core.Test/Requests/ResponseHandlerTests.cs
```diff
@@ -15,7 +15,7 @@ namespace Microsoft.Graph.DotnetCore.Core.Test.Requests
 
     public class ResponseHandlerTests
     {
-        [Fact]
+        [Fact(Skip = "Service Library needs to support System.Text.Json Attributes")]
```
ddyett: > [Fact(Skip = "Service Library needs to support System.Text.Json Attributes")] [](start = 7, length = 78)

is turning this back on tracked by an issue?

andrueastman: Yes. This is tracked with this issue https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/issues/80

src/Microsoft.Graph.Core/Serialization/DateTimeOffsetConverter.cs
andrueastman: Resolved

tests/Microsoft.Graph.DotnetCore.Core.Test/Exceptions/ErrorTests.cs
andrueastman: Resolved

PR #108, opened 2020-04-01
DeltaResponseHandler with System.Text.Json

andrueastman:
This PR migrates the DeltaResponseHandler to use System.Text.Json.

As a result the Newtonsoft dependency is now dropped from the project. 😄 

This PR also : -
- Adds a number of System.Text.Json compatible Test Models. 
This is done to drop the service library dependency that uses Newtonsoft causing a cyclic dependency on Newtonsoft and also to enable validation. (The models whould ideally be dropped once we have a compatible service library)
- Enables running of tests in the pipeline

### DIFF
--- a/src/Microsoft.Graph.Core/Requests/DeltaResponseHandler.cs
+++ b/src/Microsoft.Graph.Core/Requests/DeltaResponseHandler.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Graph
         /// <typeparam name="T">The type to return</typeparam>
         /// <param name="response">The HttpResponseMessage to handle</param>
         /// <returns></returns>
-        public async Task<T>  HandleResponse<T>(HttpResponseMessage response)
+        public async Task<T> HandleResponse<T>(HttpResponseMessage response)
         {
             if (response.Content != null)
             {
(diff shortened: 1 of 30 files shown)


src/Microsoft.Graph.Core/Requests/DeltaResponseHandler.cs
```diff
@@ -146,43 +145,107 @@ private async Task<JObject> DiscoverChangedProperties(JObject responseItem)
...
+                    string newJsonElement = this.serializer.SerializeObject(newItem);
+                    using (var newJsonDocument = JsonDocument.Parse(newJsonElement))
+                    {
+                        newJsonDocument.RootElement.WriteTo(utf8JsonWriter);// write the object
+                    }
                 }
+
+                utf8JsonWriter.WriteEndObject();//write end of object
+                //clean up
+                utf8JsonWriter.Flush();
```
MIchaelMainer, line 245: No need to manually Flush for UTF8JsonWriter. [Dispose](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Text.Json/src/System/Text/Json/Writer/Utf8JsonWriter.cs) already does this for you.

Flushing is funnier with [MemoryStream as it does nothing](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/IO/MemoryStream.cs#L176).

tests/Microsoft.Graph.DotnetCore.Core.Test/TestModels/ServiceModels/TestEmailAddress.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Microsoft.Graph.DotnetCore.Core.Test.TestModels.ServiceModels
+{
+    using System.Collections.Generic;
+    using System.Text.Json.Serialization;
+
+    /// <summary>
+    /// The type TestEmailAddress.
+    /// </summary>
+    [JsonConverter(typeof(DerivedTypeConverter<TestEmailAddress>))]
```
MIchaelMainer, line 13: EmailAddress shouldn't have a DerivedTypeConverter since it isn't derived.

src/Microsoft.Graph.Core/Requests/DeltaResponseHandler.cs
```diff
@@ -36,7 +37,7 @@ public DeltaResponseHandler()
         /// <typeparam name="T">The type to return</typeparam>
         /// <param name="response">The HttpResponseMessage to handle</param>
         /// <returns></returns>
-        public async Task<T> HandleResponse<T>(HttpResponseMessage response)
+        public async Task<T>  HandleResponse<T>(HttpResponseMessage response)
```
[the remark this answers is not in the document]
andrueastman: Resolved

tests/Microsoft.Graph.DotnetCore.Core.Test/TestModels/ServiceModels/TestEmailAddress.cs
andrueastman: Resolved

src/Microsoft.Graph.Core/Requests/DeltaResponseHandler.cs
andrueastman: Thanks for pointing this out

PR #109, opened 2020-04-03
Fix for incorrect task cancellation

### LINKED ISSUE #28: Incorrect task cancellation
I use Microsoft Graph C# SDK for managing OneDrive files. My code works with storage in parallel (max thread count < 6). For canceling an operation that is not valid anymore (user closes window or browses to another folder), I use a CancellationToken.

From time to time I receive a `Microsoft.Graph.ServiceException` with the the code `Error.Code = GraphErrorCode`.Timeout instead of `OperationCanceledException` when canceled tasks. 

Your code contains bug in `HttpProvider.SendRequestAsync` method. This method blocks propagation `OperationCanceledException` exception.

For more details please look at [StackOverflow question](https://stackoverflow.com/q/47890382/987850) and [my answer](https://stackoverflow.com/a/47900445/987850) on it.

How to fix it: You should remove `catch (TaskCanceledException exception)` block from `HttpProvider.SendRequestAsync` method.

andrueastman:
Fixes #28

- Refactored SimpleHttpProvider and HttpProvider to enable the bubbling up of the TaskCancelledException for correct task management by client apps
- Updated tests to capture this 😃 

### Other links
- https://stackoverflow.com/questions/47890382/periodic-timeout-exception-when-i-cancel-onedrive-task-that-reads-file-informati
- https://stackoverflow.com/a/47900445/987850

### DIFF
--- a/src/Microsoft.Graph.Core/Requests/HttpProvider.cs
+++ b/src/Microsoft.Graph.Core/Requests/HttpProvider.cs
@@ -244,15 +244,9 @@ namespace Microsoft.Graph
             {
                 return await this.httpClient.SendAsync(request, completionOption, cancellationToken).ConfigureAwait(false);
             }
-            catch (TaskCanceledException exception)
+            catch (OperationCanceledException)
             {
-                throw new ServiceException(
-                        new Error
-                        {
-                            Code = ErrorConstants.Codes.Timeout,
-                            Message = ErrorConstants.Messages.RequestTimedOut,
-                        },
-                        exception);
+                throw;
             }
             catch(ServiceException exception)
             {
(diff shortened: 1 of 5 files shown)


src/Microsoft.Graph.Core/Requests/HttpProvider.cs
```diff
@@ -244,15 +244,9 @@ public Task<HttpResponseMessage> SendAsync(HttpRequestMessage request)
             {
                 return await this.httpClient.SendAsync(request, completionOption, cancellationToken).ConfigureAwait(false);
             }
-            catch (TaskCanceledException exception)
+            catch (OperationCanceledException)
             {
```
ddyett, line 248: why not just remove the catch all together since nothing is done with the exception?

andrueastman: There is a final `catch` block that will catch all derived types of `Exception` and still wrap it in a ServiceException which is contrary to our desire.

PR #115, opened 2020-05-06
adds rich notifications support

baywet:
Related https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/pull/104

- adds support for decrypting and validating signature for encrypted content in notifications
- adds support for tokens validation in change notifications collections
- adds support for certificate serialization in subscriptions

### ToDo
- [ ] review interfaces names
- [ ] review extension methods names and prototypes
- [ ] test locally in webhook project
- [ ] implement unit tests

### DIFF
--- a/src/Microsoft.Graph.Core/Extensions/IDecryptableContentExtensions.cs
+++ b/src/Microsoft.Graph.Core/Extensions/IDecryptableContentExtensions.cs
@@ -8,7 +8,7 @@ using System.Threading.Tasks;
 
 namespace Microsoft.Graph
 {
-    public abstract class ChangeNotificationDecryptableContent : IDecryptableContent
+    public static class IDecryptableContentExtensions
     {
         private static readonly Lazy<int> AESInitializationVectorSize = new Lazy<int>(() =>
         {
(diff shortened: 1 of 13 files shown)


src/Microsoft.Graph.Core/Extensions/IDecryptableContentExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+            }
+        });
+
+        /// <summary>
+        /// Validates the signature and decrypted content attached with the notification.
+        /// </summary>
+        /// <typeparam name="T">Type to deserialize the data to.</typeparam>
+        /// <param name="certificateProvider">Certificate provider to decrypt the content. The first parameter is the certificate ID provided when creating the subscription. The second is the certificate thumbprint. The certificate WILL be disposed at the end of decryption.</param>
+        /// <returns>Decrypted content as the provided type.</returns>
+        public static async Task<T> Decrypt<T>(this IDecryptableContent encryptedContent, Func<string, string, Task<X509Certificate2>> certificateProvider)
```
baywet, line 27: What do you think of this "callback" approach? Would you rather have an interface defined?

src/Microsoft.Graph.Core/Extensions/IDecryptableContentExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+                return provider.LegalBlockSizes[0].MinSize;
+            }
+        });
+
+        /// <summary>
+        /// Validates the signature and decrypted content attached with the notification.
+        /// </summary>
+        /// <typeparam name="T">Type to deserialize the data to.</typeparam>
+        /// <param name="certificateProvider">Certificate provider to decrypt the content. The first parameter is the certificate ID provided when creating the subscription. The second is the certificate thumbprint. The certificate WILL be disposed at the end of decryption.</param>
+        /// <returns>Decrypted content as the provided type.</returns>
```
baywet, line 26: it feels that this generic type should have more constraints. Basically the type should be one of the types available in the SDK or a collection of a type available in the SDK. Let me know what you think

MIchaelMainer: One challenge here I see is how will the dev know what type is the type parameter. If the listener URL only receives notifications for a specific type, it will work fine. If the listener is used for subscriptions that target different types, they won't be able specify the type.  We could:
* leave it to dev to maintain a mapping of subscriptionId to the type it returns so they can use that type parameter for `Decrypt`. OR
* We add Entity (base for all Graph entities) partial to core, and provide something like we have in [DerivedTypeConverter](https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/dev/src/Microsoft.Graph.Core/Serialization/DerivedTypeConverter.cs) to deserialize to the type described in odata.type. It is deserialized into a descendant of Entity, and should have odata.type set on it.

src/Microsoft.Graph.Core/Extensions/IDecryptableContentExtensions.cs
MIchaelMainer: I prefer delegate approach like you've specified.

src/Microsoft.Graph.Core/Extensions/IEncryptableSubscriptionExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Microsoft.Graph
+{
+    public static class IEncryptableSubscriptionExtensions
+    {
+        /// <summary>
+        /// Adds the encryption certificate information for change notifications with resource data to the subscription creation information.
+        /// </summary>
+        /// <param name="certificate">Certificate to use for encryption</param>
+        public static void AddEncryptionCertificate(this IEncryptableSubscription subscription, X509Certificate2 certificate)
```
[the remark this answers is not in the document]
baywet:
```suggestion
        public static void AddPublicEncryptionCertificate(this IEncryptableSubscription subscription, X509Certificate2 certificate)
```

src/Microsoft.Graph.Core/Extensions/IEncryptableSubscriptionExtensions.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.Graph
+{
+    public static class IEncryptableSubscriptionExtensions
+    {
+        /// <summary>
+        /// Adds the encryption certificate information for change notifications with resource data to the subscription creation information.
```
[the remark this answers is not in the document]
baywet:
```suggestion
        /// Adds the public encryption certificate information for change notifications with resource data to the subscription creation information.
```

src/Microsoft.Graph.Core/Extensions/IDecryptableContentExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+        public static async Task<T> Decrypt<T>(this IDecryptableContent encryptedContent, Func<string, string, Task<X509Certificate2>> certificateProvider)
+        {
+            return JsonConvert.DeserializeObject<T>(await encryptedContent.Decrypt(certificateProvider));
+        }
+        /// <summary>
+        /// Validates the signature and decrypted content attached with the notification.
```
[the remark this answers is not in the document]
baywet:
```suggestion
        /// Validates the signature and decrypted content attached with the notification.
        /// https://docs.microsoft.com/en-us/graph/webhooks-with-resource-data#decrypting-resource-data-from-change-notifications 
```

src/Microsoft.Graph.Core/Extensions/IDecryptableContentExtensions.cs
baywet: or maybe we keep a very simple constraint like where T : class, new ?

MIchaelMainer: class makes sense, but not new. If a content data entity is a heterogenous collection with a common base type, we'd want to be able to specify the base type.

src/Microsoft.Graph.Core/Extensions/IDecryptableContentExtensions.cs
```diff
@@ -0,0 +1,86 @@
...
+            try
+            {
+                using (var cryptoServiceProvider = new AesCryptoServiceProvider
+                {
+                    Mode = CipherMode.CBC,
+                    Padding = PaddingMode.PKCS7,
+                    Key = key
+                })
+                {
+                    var numArray = new byte[AESInitializationVectorSize.Value / 8];
```
peombwa, line 68: We can get initialization vector size by getting the AES minimum legal block size (128) from the instantiated `cryptoServiceProvider` and dividing it by 8. This removes the need for the lazy initialization at the top of the class.
```suggestion
                    var numArray = new byte[cryptoServiceProvider.LegalBlockSizes[0].MinSize / 8];
```

baywet: thanks. Actually it doesn't make sense to rely on the min size as the "spec" requires the IV to be 16 bytes

baywet:
closing in favor of #117

PR #117, opened 2020-05-13
Feature/rich notifications to 2.0

baywet:
Related https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/pull/104
replaces #115 

- adds support for decrypting and validating signature for encrypted content in notifications
- adds support for tokens validation in change notifications collections
- adds support for certificate serialization in subscriptions

### ToDo
- [x] review interfaces names
- [x] review extension methods names and prototypes
- [ ] test locally in webhook project
- [ ] implement unit tests

### DIFF
--- a/src/Microsoft.Graph.Core/Extensions/IDecryptableContentExtensions.cs
+++ b/src/Microsoft.Graph.Core/Extensions/IDecryptableContentExtensions.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Graph
         /// <typeparam name="T">Type to deserialize the data to.</typeparam>
         /// <param name="certificateProvider">Certificate provider to decrypt the content. The first parameter is the certificate ID provided when creating the subscription. The second is the certificate thumbprint. The certificate WILL be disposed at the end of decryption.</param>
         /// <returns>Decrypted content as the provided type.</returns>
-        public static async Task<T> Decrypt<T>(this IDecryptableContent encryptedContent, Func<string, string, Task<X509Certificate2>> certificateProvider)
+        public static async Task<T> Decrypt<T>(this IDecryptableContent encryptedContent, Func<string, string, Task<X509Certificate2>> certificateProvider) where T : class
         {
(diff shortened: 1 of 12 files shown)


baywet:
@peombwa @MIchaelMainer I cherry-picked all my changes to this new branch based on feature/2.0 to avoid bringing other things with my PR.
Thanks for calling out that the AES provider was only available for dotnet standard 2.0 Peter!

Can you guys give it one last look before I start "implementing" in beta and testing in my project please?

Also, csproj. Right now my dependency `Microsoft.IdentityModel.Protocols.OpenIdConnect` is only added for targets 461 and standard2.0. It looks to me that I should be able to add it as a "non-conditional" reference instead as monoandroid80 and xamarin.ios10 both seems to support dotnetstandard2.0 but I wanted to double-check with you before making any decision?

MIchaelMainer:
I think you will be blocked at this point as we have to change our generation to support feature/2.0 in the service libraries. We are switching from newtonsoft to system.text.json. We will need to change the generated models to work with the feature/2.0 core.

baywet:
Is there a branch on beta SDK that's contains these changes with work in progress I could use for my testing? Or is it all still to be started?

baywet:
Alternatively we could bring the aescryptoserviceprovider from the library in the main version if we were ok with bumping from standard 1.1 to 1.3 https://www.nuget.org/packages/System.Security.Cryptography.Csp/

MIchaelMainer:
We haven't started on the beta SDK generation work. We are going to need make some changes to templates to support converters so we can maintain our API signatures.

baywet:
> Alternatively we could bring the aescryptoserviceprovider from the library in the main version if we were ok with bumping from standard 1.1 to 1.3 https://www.nuget.org/packages/System.Security.Cryptography.Csp/

I've tried to go this way but that on it's own would also require an upgrade from 4.5 to 4.6. Also Microsoft.IdentityModel.Protocols.OpenIdConnect requires dotnet standard 2.0. I guess I'll have to wait :)

andrueastman:
Closing this PR for now so that we may complete it via #269

PR #140, opened 2020-10-06
1.22 release

MIchaelMainer:


src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
```diff
@@ -22,6 +22,7 @@
     <VersionPrefix>1.22.0</VersionPrefix>
```
baywet, line 22: @MIchaelMainer aren't we missing a version bump here?

MIchaelMainer: @baywet Thank you for checking! [1.22.0 has only been released as a preview](https://www.nuget.org/packages/Microsoft.Graph.core). The version story goes like this:
1) When a new feature is added, we `feature` branch from `dev`, and bump minor version in the initial commit. So, in this case, we manually bumped to **1.22.0**.
2) When we open a PR from `feature` into `dev`, there is validation that the version has been bumped by querying the NuGet feed. If **1.22.0** has not been published, the PR is valid.
3) When the PR is commited into `dev`, the pipeline queries the Nuget feed to determine whether there has already been a preview build or not. If there isn't, it adds the preview suffix. If a preview build has already been released, it adds the preview suffix and increments the preview build number. We never check this in. 
4) The same pipeline builds and signs a preview release NuGet which gets automatically picked up for release.
5) Now here we are, wanting to release this for production. We've already incremented manually in step 1 and have validated that we incremented it. The version is set for release.

baywet: thanks for the explanation. As far as I can tell from what you've described, nothing has incremented the version in master, or merged the incrementation yet. So the version number should be marked as a change in the diff no?

MIchaelMainer: You are correct. There was supposed to be a previous 1.22 release. I don't see a release run and it obviously wasn't published. I'll see what happened this time around.

baywet: thanks for clarifying

baywet (approved):

zengin (approved):

PR #158, opened 2020-10-12
corrected DefaultMaxSliceSize value

schmitch:
default max slice is not a multiple of 320 and also the header suggests 5 MiB instead of 4

> /// If less than 0, default value of 5 MiB is used. .</param>


andrueastman:
@schmitch 
Thanks for taking the time to raise this issue. 
I have captured the change in #159 so that the build can pass. I will close this PR for now. 
Thanks again for raising this.

PR #159, opened 2020-10-13
Fix DefaultMaxSliceSize in LargeFileUpload task

### LINKED ISSUE #120: Exception in LargeFileUploadTask ctor

### Expected behavior
https://microsoftgraph.github.io/msgraph-sdk-design/tasks/FileUploadTask.html

If you do a quick find on this page, you'll see that it mentions that the default slice size is 5 MB.  I see this number in multiple places, whether it's code or docs.

I'd expect that if I don't give the `LargeFileUploadTask` ctor a value for `maxSliceSize`, it would use 5 MiB.

### Actual behavior
The default is 4 MiB, which throws an exception, because it's not fully divisible by 320 KiB.

```csharp
var uploadSession = await _client.Drives[_driveID].Items[_driveItemID].ItemWithPath(fileName).CreateUploadSession().Request().PostAsync();
var largeFileUploadTask = new LargeFileUploadTask<DriveItem>(uploadSession, ms);
```

https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/b2d9d7ce3073aabd8a65b7157f49ced0c9e28b27/src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs#L16-L19

https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/b2d9d7ce3073aabd8a65b7157f49ced0c9e28b27/src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs#L47-L51

Since 5 MiB is divisible by 320 KiB, can we just change the constant in the class?

_Side Question_
Should we be using `LargeFileUploadTask` now or `ChunkedUploadProvider`? The link I gave above for docs suggests `LargeFileUploadTask`.

andrueastman:
Fixes #120

- This PR fixes the exception thrown when creating a LargeFileUpload without specifying the slice size since the default value is not a multiple of 320KB

### Other links
- https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/73
- https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/158

### DIFF
--- a/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
+++ b/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
@@ -22,7 +22,7 @@
     <VersionPrefix>1.23.0</VersionPrefix>
     <VersionSuffix></VersionSuffix>
     <PackageReleaseNotes>
-      - Fix for #120: Exception thrown when slice size is not specified for LargeFileUploadTask 
+    - Fix for #120: Exception thrown when slice size is not specified for LargeFileUploadTask 
     </PackageReleaseNotes>
   </PropertyGroup>
   <!--We manually configure LanguageTargets for Xamarin due to .Net SDK TFMs limitation https://github.com/dotnet/sdk/issues/491 -->
--- a/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
+++ b/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
@@ -19,10 +19,10 @@
     <DelaySign>false</DelaySign>
     <AssemblyOriginatorKeyFile>35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
-    <VersionPrefix>1.22.0</VersionPrefix>
+    <VersionPrefix>1.23.0</VersionPrefix>
     <VersionSuffix></VersionSuffix>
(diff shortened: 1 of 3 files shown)


src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
```diff
@@ -19,10 +19,10 @@
     <DelaySign>false</DelaySign>
     <AssemblyOriginatorKeyFile>35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
-    <VersionPrefix>1.22.0</VersionPrefix>
+    <VersionPrefix>1.23.0</VersionPrefix>
     <VersionSuffix></VersionSuffix>
-    <PackageReleaseNotes>- BaseRequestBuilder is now an IBaseRequestBuilder.
-- Requests that return an OData primitive now work as expected. 
+    <PackageReleaseNotes>
+      - Fix for #120: Exception thrown when slice size is not specified for LargeFileUploadTask
```
MIchaelMainer, line 25: nit: Align to the left margin as the tab will make it look like https://www.nuget.org/packages/Microsoft.Graph.Core/1.21.0

MIchaelMainer (approved):

src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
andrueastman: Thanks Michael. Actioned and resolved

MIchaelMainer (approved):

PR #169, opened 2020-11-02
Add support for content type with parameters to BaseRequest

### LINKED ISSUE #165: BaseRequest ContentType does not support media types with parameters
#50 # Expected behavior

I expect the ContentType header of a request derived from BaseRequest to support the [media type parameters as defined in the HTTP specification.](https://tools.ietf.org/html/rfc2616#section-3.7)

### Actual behavior

Setting the ContentType property to a value that contains , for example "application/json;odata=verbose", throws an exception:

> The format of value 'application/json; odata=verbose' is invalid.

- Instantiate a class derived from BaseRequest
- Set the ContentType property to "application/json; odata=verbose"
- Call a method that will invoke the SendRequestAsync method

### Investigation/Proposed resolution

Parsing the ContentType property (a string) instead of passing it to the constructor of the MediaTypeHeaderValue class will properly handle content type strings that contain parameters (defined as values after a semicolon in the string).

Change [the line that sets the ContentType header](https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/1f528e15953fae6da9fb3f971e45bbfdb313d94f/src/Microsoft.Graph.Core/Requests/BaseRequest.cs#L259) of the request to:

`request.Content.Headers.ContentType = MediaTypeHeaderValue.Parse(this.ContentType);`

Related to #160  
[AB#6497](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6497)

andrueastman:
Fixes #165 

This PR adds support to Base Request for content types with parameters so that content types  like `application/json; odata=verbose` do not throw a Format Exception.

Test has been added to validate this change.

### DIFF
--- a/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
+++ b/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
@@ -24,6 +24,7 @@
     <PackageReleaseNotes>
 - Fix for #120: Exception thrown when slice size is not specified for LargeFileUploadTask
 - Fix for #113: NullReferenceException thrown when error response does not have content type header is empty.
+- Fix for #165: Add support for content types with parameters to BaseRequest
     </PackageReleaseNotes>
   </PropertyGroup>
   <!--We manually configure LanguageTargets for Xamarin due to .Net SDK TFMs limitation https://github.com/dotnet/sdk/issues/491 -->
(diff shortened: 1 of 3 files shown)


MIchaelMainer (approved):
Looks good

PR #183, opened 2020-12-08
Support composable functions from the generated service library 

### LINKED ISSUE #175: Update OData method support for setting parameters on composable functions
Parameters for functions aren't set on the URL until `.Request()` is called on the [BaseFunctionMethodRequestBuilder](https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/dc8ae2e7cc49a8f2240d2cba707ef52044ecdfa6/src/Microsoft.Graph.Core/Requests/BaseFunctionMethodRequestBuilder.cs#L66). This was fine before composable function support. With composable functions, we need to update the URL sooner or else we won't be able to pass in parameters to functions that are not the final function. This means building function parameters in the generated request builders and not in BaseFunctionMethodRequestBuilder. 

So, we will:
* remove the line66 where parameters are added in BaseFunctionMethodRequestBuilder. This means that we need to generate MethodRequestBuilder.cs.tt based request builders that will set the parameters on the URL.
* add a `protected GetParametersString()` that the generated request builders will use *after* set the parameters so that we get the fully formed parameter string for the function request builder. The generated request builder will append the parameter string to the *requestUrl`.

After inspecting this code file, it appears that we should:
* remove `passParametersInQueryString` as it is neither used in this repo nor in msgraph-sdk-dotnet. It defaults to false. Be careful of dependencies based on this value.
* Check any query options and validate that they are only [$filter or $orderby per spec](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_InvokingaFunction).
* Support the passing of both HeaderOption and QueryOption.

These changes will need to published first so we can validate the generated changes and publish them on the same day. We shouldn't list these changes on NuGet until after the generated library (v1.0 and beta) is published. This will be a breaking change for older versions of the generated libraries, so we will need to make sure we set the minimum dep version for the generated library to this version of core.

MIchaelMainer:
Fixes #175 

- Validate function query options.
- Remove the setting of parameters in Request. This is now done in `SetFunctionParameters()` in the generated request builders.
- Removed unused code. passParametersInQueryString was never set (always true) and the query options were only ever set via CreateRequest(). Definitely need another set of eyes on this. Not found in https://github.com/microsoftgraph/msgraph-sdk-dotnet/search?q=passParametersInQueryString or https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/search?q=passParametersInQueryString
- We want composable functions to always emit with parenthesis to distinguish from cast syntax. This is a change as functions without parameters weren't emitting parens. `SetFunctionParameters()` has to be called for each function request builder cstor. Note that there is some risk that the parens are not supported since this is new but ad hoc tests haven't indicated this to be the case.
- Parameter string is now set before the Request method is called. The generator must now support `SetFunctionParameters()` to set parameters on a composable function segment..

### Other links
- https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/316
- https://github.com/microsoftgraph/msgraph-sdk-dotnet/pull/848


src/Microsoft.Graph.Core/Requests/BaseFunctionMethodRequestBuilder.cs
```diff
@@ -121,5 +116,17 @@ private string EscapeStringValue(string value)
             // Per OData spec, single quotes within a string must be escaped with a second single quote.
             return value.Replace("'", "''");
         }
+
+        /// <summary>
+        /// Set the parameter string on the request URL. MUST be used after all 
+        /// parameters have been set in the generated request builder. Used for
+        /// each section. Adds an empty () if no parameters are present.
+        /// </summary>
+        protected void SetParameterString()
```
nikithauc, line 125: Can a different name be used for this method? Example - > BuildRequestURL() or something not similar to `SetParameter()`.

tests/Microsoft.Graph.DotnetCore.Core.Test/Requests/BaseRequestBuilderTests.cs
```diff
@@ -32,5 +37,107 @@ public void AppendSegmentToRequestUrl()
...
             Assert.Equal(string.Join("/", requestUrl, newUrlSegment), appendedUrl);
         }
+
+        /// <summary>
+        /// Test that: 1) composable functions have all parameters set, and 2) query options are not yet applied to the URL.
+        /// </summary>
+        [Fact]
+        public void ComposableFunctionTest()
+        {
+            var requestUrl = "https://localhost:443/microsoft.graph.composablefunction0";
```
nikithauc, line 47: requestURL, client can be stated as constants and reused in multiple tests.

src/Microsoft.Graph.Core/Requests/BaseFunctionMethodRequestBuilder.cs
```diff
@@ -48,25 +44,21 @@ public abstract class BaseFunctionMethodRequestBuilder<T> : BaseRequestBuilder w
...
-                }
-                else
+                var unexpectedQueryOptions = options.OfType<QueryOption>().Any<QueryOption>(qp => 
+                                                                                 qp.Name.ToLower() != "filter" && 
+                                                                                 qp.Name.ToLower() != "orderby");
+
+                if (unexpectedQueryOptions)
                 {
-                    options.ToList().AddRange(this._queryOptions);
+                    throw new ArgumentException("You can only use filter and orderby query options with this function.");
```
nikithauc, line 57: 1. Is this a breaking change? Are all query options being accepted before this change goes in? 
2. It would be helpful if this is stated in the summary of the function.

src/Microsoft.Graph.Core/Requests/BaseFunctionMethodRequestBuilder.cs
```diff
@@ -121,5 +116,17 @@ private string EscapeStringValue(string value)
...
+
+        /// <summary>
+        /// Set the parameter string on the request URL. MUST be used after all 
+        /// parameters have been set in the generated request builder. Used for
+        /// each section. Adds an empty () if no parameters are present.
+        /// </summary>
+        protected void SetParameterString()
+        { 
+            this.RequestUrl = string.Format("{0}({1})", 
+                                            this.RequestUrl,
```
nikithauc, line 128: What if `this.RequestURL` is null or  empty?

tests/Microsoft.Graph.DotnetCore.Core.Test/TestModels/ComposableFunctionRequestBuilder0.cs
```diff
@@ -0,0 +1,46 @@
...
+        protected override IBaseRequest CreateRequest(string functionUrl, IEnumerable<Option> options)
+        {
+            // Don't need this right now. Can implement if we need it.
+            throw new NotImplementedException();
+        }
+
+        public ComposableFunctionRequestBuilder1 RequestBuilder1(string anotherValue)
+        {
+            return new ComposableFunctionRequestBuilder1(
+            this.AppendSegmentToRequestUrl("microsoft.graph.composablefunction1"),
```
andrueastman, line 41: Nit: Indentation

tests/Microsoft.Graph.DotnetCore.Core.Test/Requests/BaseRequestBuilderTests.cs
```diff
@@ -32,5 +37,107 @@ public void AppendSegmentToRequestUrl()
...
+
+            // Create the composed function request builders
+            var composableFunctionRequestBuilder0 = new ComposableFunctionRequestBuilder0(requestUrl, client);
+            var composedRequestUrl = composableFunctionRequestBuilder0.RequestBuilder1(parameter_second_function).Request().RequestUrl;
+
+            var expected = @"https://localhost:443/microsoft.graph.composablefunction0()/microsoft.graph.composablefunction1(anotherValue='test-value')";
+
+            Assert.Equal(expected, composedRequestUrl);
+        }
+
```
andrueastman, line 102: I think it would be great to add a test that validates the final expected request url when we have the Query options (filter and orderby)present as well. 
The current tests just validate that we don't throw an error when we have them

andrueastman (commented):
👍 on the tests.

src/Microsoft.Graph.Core/Requests/BaseFunctionMethodRequestBuilder.cs
```diff
@@ -48,25 +44,21 @@ public abstract class BaseFunctionMethodRequestBuilder<T> : BaseRequestBuilder w
...
+            // Validate the query options according to 
+            // http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_InvokingaFunction
+            if (options != null)
             {
-                if (options == null)
-                {
-                    options = this._queryOptions;
-                }
-                else
+                var unexpectedQueryOptions = options.OfType<QueryOption>().Any<QueryOption>(qp =>
```
zengin, line 51: Second type annotation is redundant.
```suggestion
                var unexpectedQueryOptions = options.OfType<QueryOption>().Any(qp => 
```

src/Microsoft.Graph.Core/Requests/BaseFunctionMethodRequestBuilder.cs
```diff
@@ -48,25 +44,21 @@ public abstract class BaseFunctionMethodRequestBuilder<T> : BaseRequestBuilder w
...
+            if (options != null)
             {
-                if (options == null)
-                {
-                    options = this._queryOptions;
-                }
-                else
+                var unexpectedQueryOptions = options.OfType<QueryOption>().Any<QueryOption>(qp => 
+                                                                                 qp.Name.ToLower() != "filter" && 
+                                                                                 qp.Name.ToLower() != "orderby");
```
zengin, line 53: This usually doesn't apply to English only text, but it is good practice to always normalize strings to uppercase: https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2015/code-quality/ca1308-normalize-strings-to-uppercase?view=vs-2015&redirectedfrom=MSDN

src/Microsoft.Graph.Core/Requests/BaseFunctionMethodRequestBuilder.cs
zengin: I think this is moving the error from service side to client side, i.e. an early failure in the stack. Technically breaking, but I wouldn't count it as something that requires a major version bump.

tests/Microsoft.Graph.DotnetCore.Core.Test/Requests/BaseRequestBuilderTests.cs
zengin: Could you also add a test that validates the concatenation of comma separated parameters?

zengin (dismissed):

src/Microsoft.Graph.Core/Requests/BaseFunctionMethodRequestBuilder.cs
```diff
@@ -80,6 +72,16 @@ public T Request(IEnumerable<Option> options = null)
         /// <returns>A string representing the parameter for an OData method call.</returns>
         protected void SetParameter(string name, object value, bool nullable)
         {
+            if (name == string.Empty)
```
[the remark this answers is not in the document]
nikithauc: string.IsNullOrWhiteSpace would be better may be.

src/Microsoft.Graph.Core/Requests/BaseFunctionMethodRequestBuilder.cs
nikithauc: Some name that indicates the concatenation/ building of the complete URL.

MIchaelMainer: This won't build the request URL. It only adds the current request builders parameters to the request URL. The request URL can have query strings added later at the end of the request builder chain. 

I'm comfortable with them having similar names as the functionality is related. Would SetParameterStringOnUrl() be clearer?

src/Microsoft.Graph.Core/Requests/BaseFunctionMethodRequestBuilder.cs
MIchaelMainer: Good question.

Composable OData functions are bound to a type. This means we'd have to have a request builder for the bound type that would call this function request builder. The request builder for the bound type would have a request URL that would be passed into the cstor of this type. If `this.RequestURL` is null or empty, we'd have problems elsewhere.

src/Microsoft.Graph.Core/Requests/BaseFunctionMethodRequestBuilder.cs
MIchaelMainer: Thank you. Fixed.

src/Microsoft.Graph.Core/Requests/BaseFunctionMethodRequestBuilder.cs
MIchaelMainer: Done

src/Microsoft.Graph.Core/Requests/BaseFunctionMethodRequestBuilder.cs
```diff
@@ -80,6 +72,16 @@ public T Request(IEnumerable<Option> options = null)
...
+            { 
+                throw new ClientException(
+                    new Error
+                    {
+                        Code = "invalidRequest",
+                        Message = "Parameter name must not be an empty string."
```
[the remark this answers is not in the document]
MIchaelMainer: Fixed

src/Microsoft.Graph.Core/Requests/BaseFunctionMethodRequestBuilder.cs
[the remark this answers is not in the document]
MIchaelMainer: Fixed

tests/Microsoft.Graph.DotnetCore.Core.Test/Requests/BaseRequestBuilderTests.cs
MIchaelMainer: Good idea. Done.

tests/Microsoft.Graph.DotnetCore.Core.Test/TestModels/ComposableFunctionRequestBuilder0.cs
MIchaelMainer: Fixed

zengin (dismissed):

src/Microsoft.Graph.Core/Requests/BaseFunctionMethodRequestBuilder.cs
MIchaelMainer: 1) I haven't seen functions used with query params. With that stated, we arent a strict OData service so doing this may not be the best thing todo.  I'm going to remove this check.
2) Will do. Note that this is only (so far) been called from generated classes.

tests/Microsoft.Graph.DotnetCore.Core.Test/Requests/BaseRequestBuilderTests.cs
MIchaelMainer: Fixed

zengin (approved):

andrueastman (approved):

nikithauc (approved):

baywet (approved):

src/Microsoft.Graph.Core/Requests/BaseFunctionMethodRequestBuilder.cs
MIchaelMainer: I removed this check. 1) I can't totally trust that I know whether these querystrings aren't in actual use, 2) and into the future, I don't totally trust the service to not support these.

PR #218, opened 2021-03-02
Change Method type in IBaseRequest to enum

### LINKED ISSUE #217: Change Method type from string to enum in IBaseRequest
The enum should be defined in the core library, and the value carried over all the way to the core library. This would make comparisons faster in conditions as @zengin mentioned earlier. It'd would also reduce the risk of getting a unexpected value.

_Originally posted by @baywet in https://github.com/microsoftgraph/msgraph-sdk-dotnet/pull/905#discussion_r581943533_ 
[AB#8344](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8344)

andrueastman:
This PR closes #217 

It involves the changing of the Method type from string to enum to prevent erroneous inputs and provide central references. 

This PR also replaces the inline use of string values for "application/json" and "application/octet-stream" with Constants provided in the CoreConstants file.

The aim is to use these constants in the service library using central references from the core library to prevent redefinitions of constant values.

Related 
- https://github.com/microsoftgraph/msgraph-sdk-dotnet/pull/905

### DIFF
--- a/src/Microsoft.Graph.Core/CoreConstants.cs
+++ b/src/Microsoft.Graph.Core/CoreConstants.cs
@@ -56,6 +56,9 @@ namespace Microsoft.Graph
             {
                 /// JSON content type value
                 public const string Json = "application/json";
+
+                /// Stream content type value
+                public const string Stream = "application/octet-stream";
             }
         }
 
--- a/src/Microsoft.Graph.Core/CoreConstants.cs
+++ b/src/Microsoft.Graph.Core/CoreConstants.cs
@@ -138,7 +138,28 @@ namespace Microsoft.Graph
             /// <summary>
             /// The DELETE method.
             /// </summary>
-            DELETE
+            DELETE,
+
+            /// <summary>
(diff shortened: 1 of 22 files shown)


src/Microsoft.Graph.Core/CoreConstants.cs
```diff
@@ -138,7 +141,28 @@ public enum HttpMethods
             /// <summary>
             /// The DELETE method.
             /// </summary>
-            DELETE
+            DELETE,
```
baywet, line 144: nit: move the enum to it's own file (not under the core constants class) to simplify imports

andrueastman: Thanks. This is now resolved via https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/218/commits/6010bc263ba39bd484e091c5ece90f0fc6dfc6e4

baywet (approved):

PR #223, opened 2021-03-20
Drain response content when shouldretry

MIchaelMainer:


src/Microsoft.Graph.Core/Requests/Middleware/RetryHandler.cs
```diff
@@ -78,14 +78,15 @@ private async Task<HttpResponseMessage> SendRetryAsync(HttpResponseMessage respo
             int retryCount = 0;
             TimeSpan cumulativeDelay = TimeSpan.Zero;
 
-            while (retryCount < RetryOption.MaxRetry)
+            // Drain response content to free connections. Need to perform this
+            // before retry attempt and before the TooManyRetries ServiceException.
+            if (response.Content != null)
```
zengin, line 83: now, we are not draining in the loop. I think this needs to happen after the loop, but before the ServiceException.

zengin (approved):

PR #226, opened 2021-03-25
Update RetryHandler.cs - drain response before exception

MIchaelMainer:
Same change as https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/223


zengin:
I was going to suggest taking the fixes from master directly, but it looks like these two branches diverged a lot. That may cause issues while merging feature/2.0 to master.

zengin (approved):

PR #233, opened 2021-03-29
Logging handler guidance

andrueastman:
This PR is aimed at providing a central, consistent, and quick reference/guidance for users of the SDK who wish to log requests made by the SDK. 
This is especially helpful if more information is required when we need to isolate/identify issues that may be going wrong with the SDK,

This is aimed providing a to give users if this is needed.

### DIFF
--- a/docs/logging-requests.md
+++ b/docs/logging-requests.md
@@ -1,13 +1,17 @@
 # Logging requests in the Microsoft Graph .NET Client Library
 
-The SDK does not log of request/response infomation out of the box. If you wish to log request and response information performed by the sdk there are two options. These are :-
+The SDK does not log of request/response infomation out of the box. If you wish to log request and response information performed by the sdk, there are currently two ways do this. These are :-
 
-a. Implement a LoggingHandler and add it to the list of handlers.
-b. Take advantage of the fact that .NET has an inbuilt [DiagnosticHandler](https://github.com/dotnet/runtime/blob/4f8be1992cc79807d377afce640f219d4caffb5b/src/libraries/System.Net.Http/src/System/Net/Http/DiagnosticsHandler.cs#L15) in the [HttpClientHandler](https://github.com/dotnet/runtime/blob/4f8be1992cc79807d377afce640f219d4caffb5b/src/libraries/System.Net.Http/src/System/Net/Http/HttpClientHandler.cs#L33) which is used by default by the SDK.
+    a. Implement a LoggingHandler and add it to the list of handlers.
+    b. Take advantage of OpenTelemetry.NET's instrumentation of HttpClient
 
 ## a. Implementing a LoggingHandler
 
-You can implement a logging handler that looks something like this. (The example below simply logs a statement to show that the request has been sent out and the response has come back)
+### Step 1: Implement a HttpClient Message Handler
+
+You can read more about HttpClient Message Handlers [here](https://docs.microsoft.com/en-us/aspnet/web-api/overview/advanced/httpclient-message-handlers).
+
+The implemented logging handler will look something like the example below. The example simply logs a statement to show that the request has been sent out and the response has come back.
 
 ```cs
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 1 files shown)


docs/logging-requests.md
```diff
@@ -0,0 +1,175 @@
+# Logging requests in the Microsoft Graph .NET Client Library
+
+The SDK does not log of request/response infomation out of the box. If you wish to log request and response information performed by the sdk, there are currently two ways do this. These are :-
```
baywet, line 3:
```suggestion
The SDK does not log of request/response infomation out of the box. If you wish to log request and response information performed by the sdk, there are currently two ways do this. These are :
```

docs/logging-requests.md
```diff
@@ -0,0 +1,175 @@
...
+```
+
+### Step 2: Enable HTTP Instrumentation in your application.
+
+This example also sets up the OpenTelemetry Console exporter, which requires adding the package
+[`OpenTelemetry.Exporter.Console`](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.Console/README.md)
+to the application.
+
+```cs
+using var tracerProvider = Sdk.CreateTracerProviderBuilder()
```
baywet, line 90: where does this Sdk symbol come from?

baywet (approved):

docs/logging-requests.md
andrueastman: This comes from the `OpenTelemetry.Instrumentation.Http` package to be installed in Step 1

baywet: ok, thanks for the information. It's "strange" they chose such a generic name for the entry point.
As the snippet doesn't feature usings, maybe add a comment that says Sdk comes from "namespace name"? 
This is a question that comes often in samples "where does this thing come from?"

andrueastman: That is valid. Will action and resolve.

baywet (approved):

andrueastman:
Thanks for going through this @baywet!

docs/logging-requests.md
```diff
@@ -0,0 +1,178 @@
...
+        /// <summary>
+        /// Sends a HTTP request.
+        /// </summary>
+        /// <param name="httpRequest">The <see cref="HttpRequestMessage"/> to be sent.</param>
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/> for the request.</param>
+        /// <returns></returns>
+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage httpRequest, CancellationToken cancellationToken)
+        {
+            _logger.LogInformation("Sending Graph request", httpRequest);// log the request before it goes out.
+            HttpResponseMessage response = await base.SendAsync(httpRequest, cancellationToken);
```
MIchaelMainer, line 41: We really should specify the requirement that we always want to call `base.SendAsync`. 

In terms of best practices, in case of an exception in the logger, we always finally drain the response.

docs/logging-requests.md
```diff
@@ -0,0 +1,178 @@
...
+            _logger.LogInformation("Received Graph response", response);// log the response as it comes back.
+            return response;
+        }
+    }
+}
+```
+
+### Step 2: Add the Handler to GraphServiceClient's list of handlers
+
+Once the logging handler is implemented. You can add it to the list of handlers used by the SDK as follows.
```
MIchaelMainer, line 51: We should specify the best practice of adding it at the end of the handler list so that it captures all requests/responses regardless of what the earlier handlers added.

andrueastman: This is valid. Will address this in a separate PR.

docs/logging-requests.md
andrueastman: This is valid. Will address this in a separate PR.

docs/logging-requests.md
andrueastman: Resolved through #235

PR #234, opened 2021-03-29
Fix for culture dependent DateTimeOffset serialization.

### LINKED ISSUE #232: DateTimeOffset serialization is culture dependent in 2.0.0-preview

### Expected behavior

When creating a `Subscription`, which includes the required `ExpirationDateTime` fields of type `DateTimeOffset`, I expected the following field assignment to work based on the [examples in the documentation](https://docs.microsoft.com/en-us/graph/api/subscription-post-subscriptions?view=graph-rest-1.0&tabs=csharp) and elsewhere:

```c#
await _client.Subscriptions.Request()
    .AddAsync(new Subscription
    {
        ExpirationDateTime = DateTimeOffset.UtcNow.AddMinutes(5),
        ...
    });
```

### Actual behavior

The request fails with the following message: `Could not process subscription creation payload. Are all property names spelled and camelCased properly?Also are the dateTimeOffest properties in a valid internet Date and Time format?`.

Looking at the request that gets sent to the API, I noticed that the `expirationDateTime` field is formatted according to the current culture instead of using the format the API expects:

```json
{
  "expirationDateTime": "25/03/2021 18:10:53 +00:00",
  ...
}
```

After a bit of debugging, I believe the culprit is the `DateTimeOffsetConverter` used [here](https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blame/Microsoft.Graph.Core.2.0.0-preview.8.nupkg/src/Microsoft.Graph.Core/Serialization/Serializer.cs#L40), as it converts the timestamp to string [using `ToString`](https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/Microsoft.Graph.Core.2.0.0-preview.8.nupkg/src/Microsoft.Graph.Core/Serialization/DateTimeOffsetConverter.cs#L51), whose output is culture dependent.

Create a `Subscription` while using `Microsoft.Graph.Core (2.0.0-preview.8)` through `Microsoft.Graph.Beta (4.0.0-preview)`.

 
[AB#8745](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8745)

andrueastman:
This PR closes #232 

This fixes the serialization of DateTimeOffset objects to no longer be culture dependent to emit consistent values for the strings accoring to the ISO 8601-1:2019 format

### DIFF
--- a/tests/Microsoft.Graph.DotnetCore.Core.Test/Serialization/SerializerTests.cs
+++ b/tests/Microsoft.Graph.DotnetCore.Core.Test/Serialization/SerializerTests.cs
@@ -4,6 +4,7 @@
 
 namespace Microsoft.Graph.DotnetCore.Core.Test.Serialization
 {
+    using Microsoft.Graph.Core.Models;
     using Microsoft.Graph.DotnetCore.Core.Test.TestModels;
     using System;
     using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


tests/Microsoft.Graph.DotnetCore.Core.Test/Serialization/SerializerTests.cs
```diff
@@ -339,6 +340,38 @@ public void SerializeDateValue()
             Assert.Equal(expectedSerializedString, serializedString);
         }
 
+        [Fact]
+        public void SerializeDateTimeOffsetValue()
+        {
+            // Arrange
+            var dateTimeOffset = DateTimeOffset.Parse("2016-11-20T18:23:45.9356913+00:00");
```
baywet, line 347: nit: you should also have unit tests with offsets, our CI runs on machines configured on UTC, which often leads to false negatives

baywet (approved):

tests/Microsoft.Graph.DotnetCore.Core.Test/Serialization/SerializerTests.cs
andrueastman: Resolved via https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/234/commits/a50d4a6b08a8e0944b7474f094b21b0fe314b8c8

baywet (approved):

PR #242, opened 2021-04-07
Fix for Request().WithShouldRetry is not called

### LINKED ISSUE #190: Request().WithShouldRetry is not called for failed requests.
### Expected behavior
i am trying to overcome intermittent 409 error that occurs while uploading/updating metadata of a file in SharePoint's Document library using Microsoft Graph SDK. To retry failed calls SDK provides WithMaxRetry() and WithShouldRetry() options. The MaxRetry works for error codes 429, and I am assuming that ShouldRetry delegate offers us an option to implement our own evaluation logic for the retries. Based on this assumption, I have the below code:
```
_graphServiceClientFactory.GetClient().Drives[driveId].Root.ItemWithPath(path).ListItem.Fields.Request()
                       .WithShouldRetry((delay, attempt, httpResponse) =>
                        (attempt <= 5 &&
                        (httpResponse.StatusCode == HttpStatusCode.Conflict)))
                       .UpdateAsync(new FieldValueSet { AdditionalData = dataDictionary });
```

### Actual behavior
In my test the ShouldRetry delegate is never evaluated on failed requests.  I am not sure if the above usage is correct (i could not find any reference of its usage), please suggest otherwise.

Thanks. 
[AB#7188](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7188)

andrueastman:
This PR closes #190 

When a user specifies the custom delegate `ShouldRetry` in the RetryHandlerOption.cs the delegate will not be respected if the request does not respect the conditions set out in the `ShouldRetry` private method [here](https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/ba5eb6825f9de8e12707907ade8d548314688574/src/Microsoft.Graph.Core/Requests/Middleware/RetryHandler.cs#L192).  This is in turn leads to the result that users can only filter out requests to be retried rather than filtering in more options to be retried.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/242)

### DIFF
--- a/src/Microsoft.Graph.Core/Requests/Middleware/Options/RetryHandlerOption.cs
+++ b/src/Microsoft.Graph.Core/Requests/Middleware/Options/RetryHandlerOption.cs
@@ -79,8 +79,8 @@ namespace Microsoft.Graph
 
         /// <summary>
         /// A delegate that's called to determine whether a request should be retried or not.
-        /// The delegate method should accept a delay time in seconds of, number of retry attempts and <see cref="HttpResponseMessage"/> as it's parameters and return a <see cref="bool"/>. This defaults to true
+        /// The delegate method should accept a delay time in seconds of, number of retry attempts and <see cref="HttpResponseMessage"/> as it's parameters and return a <see cref="bool"/>. This defaults to false
         /// </summary>
-        public Func<int, int, HttpResponseMessage, bool> ShouldRetry { get; set; } = (delay, attempt, response) => true;
+        public Func<int, int, HttpResponseMessage, bool> ShouldRetry { get; set; } = (delay, attempt, response) => false;
     }
 }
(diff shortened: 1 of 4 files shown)


baywet:
So if I'm following properly, users won't be able NOT to retry on a 429 with the current implementation, correct?

andrueastman:
That is a very good question. 
If they wanted to do so, they would have to do so by setting the `RetryOption.MaxRetry` option to zero.

MIchaelMainer:
Setting MaxRetry to zero would also mean not retrying on 503 and 504.

MIchaelMainer (approved):

PR #249, opened 2021-05-05
JsonPropertyName attribute to match property names

andrueastman:
This PR modifies the DerivedTypeConverter to use the string value provided in the `JsonPropertyName` attribute to match properties during deserialization. 
This was initially done using the Property name in the object and therefore properties would sometimes be put into the additional Data rather than being put into the property since the names would not match.

Related issue:
https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/issues/291

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/249)

### DIFF
--- a/src/Microsoft.Graph.Core/Serialization/DerivedTypeConverter.cs
+++ b/src/Microsoft.Graph.Core/Serialization/DerivedTypeConverter.cs
@@ -123,8 +123,12 @@ namespace Microsoft.Graph
                     // iterate through the object properties
                     foreach (var property in json.EnumerateObject())
                     {
-                        // look up the property in the object definition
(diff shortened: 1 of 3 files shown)


MIchaelMainer (approved):

MIchaelMainer:
What happens if a property is renamed in a type that is not a derived type?

andrueastman:
The serialization should be okay in that case and be consistent with this behavior since it will use the serializer's default behavior. 
Ideally, once System.Text.Json has a working equivalent of Newtonsoft.Json's `JsonConvert.PopulateObject` we should be able to invoke the default serailizer behavior as we populate the derived type as we did before. More info [here](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to?pivots=dotnet-core-3-1#populate-existing-objects)

PR #260, opened 2021-06-09
Fix for null reference on error property in some error payloads

andrueastman:
This PR closes https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/issues/100

In the event that the error response from graph has an unexpected property set in the payload as null, this causes NullReferenceException as the code tries to call the ToString method on a null value. 

The change therefore ensures that the SDK user will get the full error information and not hindered by the null reference exception in this case.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/260)

### DIFF
--- a/src/Microsoft.Graph.Core/Exceptions/Error.cs
+++ b/src/Microsoft.Graph.Core/Exceptions/Error.cs
@@ -128,7 +128,7 @@ namespace Microsoft.Graph
                 errorStringBuilder.Append(Environment.NewLine);
                 foreach (var prop in this.AdditionalData)
                 {
-                    errorStringBuilder.AppendFormat("\t{0}: {1}", prop.Key, prop.Value == null ? "null" : prop.Value.ToString());
+                    errorStringBuilder.AppendFormat("\t{0}: {1}", prop.Key, prop.Value?.ToString() ?? "null");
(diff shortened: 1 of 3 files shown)


src/Microsoft.Graph.Core/Exceptions/Error.cs
```diff
@@ -128,7 +128,7 @@ public override string ToString()
                 errorStringBuilder.Append(Environment.NewLine);
                 foreach (var prop in this.AdditionalData)
                 {
-                    errorStringBuilder.AppendFormat("\t{0}: {1}", prop.Key, prop.Value.ToString());
+                    errorStringBuilder.AppendFormat("\t{0}: {1}", prop.Key, prop.Value == null ? "null" : prop.Value.ToString());
```
baywet, line 131:
```suggestion
                    errorStringBuilder.AppendFormat("\t{0}: {1}", prop.Key, prop.Value?.ToString() ?? "null");
```

baywet (changes requested):

baywet (approved):

andrueastman:
Thanks for taking a look at this @baywet

PR #271, opened 2021-07-08
Merge feature/2.0 into dev

andrueastman:
This PR merges the feature/2.0 branch into the dev branch in preparation for releasing v2.0.0 of the SDK.

Merge conflicts will be resolved via #272 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/271)


src/Microsoft.Graph.Core/Authentication/MsalAuthenticationProviderOption.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.Graph
+{
+    using Microsoft.Identity.Client;
+    using System.Security;
+
+    /// <summary>
+    /// Options class used to configure the authentication providers.
+    /// </summary>
+    internal class MsalAuthenticationProviderOption : IAuthenticationProviderOption
```
baywet, line 13: I'd rename it ScopedAuthenticationProviderOptions since there's nothing related specifically to MSAL here.

src/Microsoft.Graph.Core/Authentication/TokenCredentials/IntegratedWindowsTokenCredential.cs
```diff
@@ -0,0 +1,138 @@
...
+    using Microsoft.Identity.Client;
+    using System;
+
+    /// <summary>
+    /// A <see cref="TokenCredential"/> implementation using MSAL.Net to acquire token by integrated windows authentication.
+    /// </summary>
+    public class IntegratedWindowsTokenCredential : TokenCredential
+    {
+
+        private static readonly List<string> WellKnownTenants = new List<string>
```
baywet, line 21:
```suggestion
        private static readonly HashSet<string> WellKnownTenants = new HashSet<string>
```

and replace any string lookup, complexity becomes o(1) instead of o(n)

src/Microsoft.Graph.Core/Authentication/TokenCredentials/IntegratedWindowsTokenCredential.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        /// <summary>
+        /// Gets an <see cref="AccessToken"/> using the provided <see cref="TokenRequestContext"/> in a synchronous fashion
+        /// </summary>
+        /// <param name="requestContext">The <see cref="TokenRequestContext"/> for the request</param>
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/> for the request</param>
+        /// <returns>An <see cref="AccessToken"/> to make requests with</returns>
+        public override AccessToken GetToken(TokenRequestContext requestContext, CancellationToken cancellationToken)
+        {
+            return GetTokenAsync(requestContext,cancellationToken).GetAwaiter().GetResult();
```
baywet, line 49: wrapping in a try catch to unwrap the TaskException and return throw the source exception instead might be required here (part of what await does besides waiting for the task to complete)

src/Microsoft.Graph.Core/Extensions/BaseRequestExtensions.cs
```diff
@@ -167,5 +167,41 @@ public static class BaseRequestExtensions
...
+        /// Sets Microsoft Graph's scopes that will be used by <see cref="IAuthenticationProvider"/> to authenticate this request
+        /// and can be used to perform incremental scope consent.
+        /// This only works with the default authentication handler and default set of Microsoft graph authentication providers.
+        /// If you use a custom authentication handler or authentication provider, you have to handle it's retrieval in your implementation.
+        /// </summary>
+        /// <param name="baseRequest">The <see cref="IBaseRequest"/>.</param>
+        /// <param name="scopes">Microsoft graph scopes used to authenticate this request.</param>
+        public static T WithScopes<T>(this T baseRequest, string[] scopes) where T : IBaseRequest
+        {
+            string authHandlerOptionKey = typeof(AuthenticationHandlerOption).ToString();
```
baywet, line 181: nameof instead?

src/Microsoft.Graph.Core/Extensions/TokenCredentialExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        /// <summary>
+        /// Gets retry after timespan from <see cref="RetryConditionHeaderValue"/>.
+        /// </summary>
+        /// <param name="tokenCredential">n <see cref="TokenCredential"/> object.</param>
+        /// <param name="serviceException">A <see cref="MsalServiceException"/> with RetryAfter header</param>
+        internal static TimeSpan GetRetryAfter(this TokenCredential tokenCredential, MsalServiceException serviceException)
+        {
+            RetryConditionHeaderValue retryAfter = serviceException.Headers?.RetryAfter;
+            TimeSpan? delay = null;
```
baywet, line 23:
```suggestion
            TimeSpan? delay = retryAfter?.Delta ?? retryAfter?.Date?.Value?.Offset;
```
and remove the if/else if blocks below

src/Microsoft.Graph.Core/Requests/BaseClient.cs
```diff
@@ -31,6 +33,25 @@ public class BaseClient : IBaseClient
...
+        /// <param name="scopes">List of scopes for the authentication context.</param>
+        /// <param name="httpProvider">The <see cref="IHttpProvider"/> for sending requests.</param>
+        public BaseClient(
+            string baseUrl,
+            TokenCredential tokenCredential,
+            IEnumerable<string> scopes = null,
+            IHttpProvider httpProvider = null
+            )
+        {
+            this.BaseUrl = baseUrl;
```
baywet, line 50: no validation on the base url?

src/Microsoft.Graph.Core/Requests/Content/BatchResponseContent.cs
```diff
@@ -55,10 +60,10 @@ public BatchResponseContent(HttpResponseMessage httpResponseMessage, ISerializer
             if (jBatchResponseObject == null)
                 return responseMessages;
 
-            if(jBatchResponseObject.TryGetValue(CoreConstants.BatchRequest.Responses, out JToken jResponses))
+            if(jBatchResponseObject.RootElement.TryGetProperty(CoreConstants.BatchRequest.Responses, out JsonElement jResponses))
```
baywet, line 63:
```suggestion
            if(jBatchResponseObject.RootElement.TryGetProperty(CoreConstants.BatchRequest.Responses, out JsonElement jResponses) && jResponses != null)
```

src/Microsoft.Graph.Core/Requests/Content/BatchResponseContent.cs
```diff
@@ -75,14 +80,18 @@ public async Task<HttpResponseMessage> GetResponseByIdAsync(string requestId)
             if (jBatchResponseObject == null)
                 return null;
 
-            JObject jResponseItem = null;
-
-            if (jBatchResponseObject.TryGetValue(CoreConstants.BatchRequest.Responses, out JToken jResponses))
+            if (jBatchResponseObject.RootElement.TryGetProperty(CoreConstants.BatchRequest.Responses, out JsonElement jResponses))
```
baywet, line 83:
```suggestion
            if (jBatchResponseObject.RootElement.TryGetProperty(CoreConstants.BatchRequest.Responses, out JsonElement jResponses) && jResponses != null)
```

src/Microsoft.Graph.Core/Requests/Content/BatchResponseContent.cs
baywet: probably need to test whether it's an array as well actually

src/Microsoft.Graph.Core/Requests/DeltaResponseHandler.cs
```diff
@@ -146,49 +145,106 @@ private async Task<JObject> DiscoverChangedProperties(JObject responseItem)
         /// <param name="changes">The list of properties returned in the response.</param>
         /// <param name="parentName">The parent object of this changed object.</param>
         /// <returns></returns>
-        private async Task GetObjectProperties(JObject changedObject, JArray changes, string parentName = "")
+        private async Task GetObjectProperties(JsonElement changedObject, List<string> changes, string parentName = "")
         {
             if (parentName != string.Empty)
```
baywet, line 150: string is null or empty

src/Microsoft.Graph.Core/Requests/DeltaResponseHandler.cs
```diff
@@ -146,49 +145,106 @@ private async Task<JObject> DiscoverChangedProperties(JObject responseItem)
...
             {
                 parentName += ".";
             }
 
-            foreach (JProperty property in changedObject.Children())
+            foreach (var property in changedObject.EnumerateObject())
```
baywet, line 155: what happens if changedObject is null?

src/Microsoft.Graph.Core/Requests/GraphResponse.cs
```diff
@@ -0,0 +1,66 @@
...
+    public class GraphResponse :IDisposable
+    {
+        /// <summary>
+        /// The GraphResponse Constructor
+        /// </summary>
+        /// <param name="iBaseRequest">The Request made for the response</param>
+        /// <param name="httpResponseMessage">The response</param>
+        public GraphResponse(IBaseRequest iBaseRequest, HttpResponseMessage httpResponseMessage)
+        {
+            this.httpResponseMessage = httpResponseMessage;
```
baywet, line 23: no validation on the passed values?

src/Microsoft.Graph.Core/Requests/HttpProvider.cs
```diff
@@ -213,7 +213,7 @@ public Task<HttpResponseMessage> SendAsync(HttpRequestMessage request)
                         }
                     }
 
-                    if (response.Content?.Headers.ContentType?.MediaType == "application/json")
+                    if (response.Content?.Headers.ContentType?.MediaType == CoreConstants.MimeTypeNames.Application.Json)
```
baywet, line 216: string Equals, ignore case

src/Microsoft.Graph.Core/Requests/SimpleHttpProvider.cs
```diff
@@ -135,7 +135,7 @@ public Task<HttpResponseMessage> SendAsync(HttpRequestMessage request)
                         }
                     }
 
-                    if (response.Content?.Headers.ContentType?.MediaType == "application/json")
+                    if (response.Content?.Headers.ContentType?.MediaType == CoreConstants.MimeTypeNames.Application.Json)
```
baywet, line 138: casing

src/Microsoft.Graph.Core/Tasks/PageIterator.cs
```diff
@@ -0,0 +1,269 @@
...
+        /// Creates the PageIterator with the results of an initial paged request. 
+        /// </summary>
+        /// <param name="client">The GraphServiceClient object used to create the NextPageRequest for a delta query.</param>
+        /// <param name="page">A generated implementation of ICollectionPage.</param>
+        /// <param name="callback">A Func delegate that processes type TEntity in the result set and should return false if the iterator should cancel processing.</param>
+        /// <param name="requestConfigurator">A Func delegate that configures the NextPageRequest</param>
+        /// <returns>A PageIterator&lt;TEntity&gt; that will process additional result pages based on the rules specified in Func&lt;TEntity,bool&gt; processPageItems</returns>
+        public static PageIterator<TEntity> CreatePageIterator(IBaseClient client, ICollectionPage<TEntity> page, Func<TEntity, bool> callback, Func<IBaseRequest, IBaseRequest> requestConfigurator = null)
+        {
+            if (page == null)
```
baywet, line 52: no test on client?

src/Microsoft.Graph.Core/Tasks/PageIterator.cs
```diff
@@ -0,0 +1,269 @@
...
+    */
+
+    /// <summary>
+    /// Use PageIterator&lt;TEntity&gt; to automatically page through result sets across multiple calls 
+    /// and process each item in the result set.
+    /// </summary>
+    /// <typeparam name="TEntity">The Microsoft Graph entity type returned in the result set.</typeparam>
+    public class PageIterator<TEntity>
+    {
+        private IBaseClient _client;
```
baywet, line 23: readonly?

src/Microsoft.Graph.Core/Tasks/PageIterator.cs
```diff
@@ -0,0 +1,269 @@
...
+
+            // Get the next page if it is available and queue the items for processing.
+            if (page.NextPageRequest != null)
+            {
+                // Call the MSGraph API to get the next page of results and set that page as the currentPage.
+                _currentPage = await (_requestConfigurator == null ? page.NextPageRequest : _requestConfigurator(page.NextPageRequest)).GetAsync(token).ConfigureAwait(false);
+                page = _currentPage;
+
+                // Add all of the items returned in the response to the queue.
+                if (_currentPage.Count > 0)
```
baywet, line 146:
```suggestion
                if (_currentPage != null && _currentPage.Count > 0)
```

baywet (changes requested):

MIchaelMainer:
Will need to review locally. Let's make sure we have an archive branch for the old version.

src/Microsoft.Graph.Core/Requests/BaseClient.cs
andrueastman: This is already done [here ](https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/8c83235d0bca749a05a78b270cdf7e869a278c49/src/Microsoft.Graph.Core/Requests/BaseClient.cs#L79)on the property setter.

src/Microsoft.Graph.Core/Requests/Content/BatchResponseContent.cs
andrueastman: JsonElement is actually a struct and is therefore not nullable. Will resolve with the check `jResponses.ValueKind == JsonValueKind.Array`

src/Microsoft.Graph.Core/Requests/Content/BatchResponseContent.cs
andrueastman: JsonElement is actually a struct and is therefore not nullable. Will resolve with the check jResponses.ValueKind == JsonValueKind.Array

src/Microsoft.Graph.Core/Requests/DeltaResponseHandler.cs
andrueastman: JsonElement is actually a struct and is actually not nullable(value type).

src/Microsoft.Graph.Core/Tasks/PageIterator.cs
andrueastman: _client is also settable via the static method `CreatePageIterator`. For it to be readonly it would only be set by the constructor.

src/Microsoft.Graph.Core/Authentication/TokenCredentials/IntegratedWindowsTokenCredential.cs
andrueastman: Isn't that the case when you expect an `AggregateException` from calling .Result directly?
https://stackoverflow.com/questions/34793684/catching-original-exception-when-using-async-synchronously

pipelines/validatePR.yml
```diff
@@ -31,7 +31,10 @@ variables:
   PROJECT_PATH: '.\src\Microsoft.Graph.Core\Microsoft.Graph.Core.csproj'
 
 pool:
-  vmImage: windows-latest
+  name: Hosted Windows 2019 with VS2019
```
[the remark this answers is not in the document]
andrueastman: Resolved via #271

andrueastman:
@baywet @MIchaelMainer 
Thanks for the taking the time and feedback on the PR. I have now resolved the comments via #274 to be merged into this.

src/Microsoft.Graph.Core/Tasks/PageIterator.cs
andrueastman: Resolved via #274

src/Microsoft.Graph.Core/Tasks/PageIterator.cs
andrueastman: Resolved via #274

src/Microsoft.Graph.Core/Requests/SimpleHttpProvider.cs
andrueastman: Resolved via #274

src/Microsoft.Graph.Core/Requests/HttpProvider.cs
andrueastman: Resolved via #274

src/Microsoft.Graph.Core/Requests/GraphResponse.cs
andrueastman: Resolved via #274

src/Microsoft.Graph.Core/Requests/DeltaResponseHandler.cs
andrueastman: Resolved via #274

src/Microsoft.Graph.Core/Extensions/TokenCredentialExtensions.cs
andrueastman: Resolved via #274

src/Microsoft.Graph.Core/Authentication/TokenCredentials/IntegratedWindowsTokenCredential.cs
baywet: interesting, I didn't know this subtilty between .Result and .GetResult()! Thanks for sharing. Resolving.

src/Microsoft.Graph.Core/Extensions/BaseRequestExtensions.cs
baywet: unresolving this one, it looks like it's been missed

baywet (approved):

src/Microsoft.Graph.Core/Extensions/BaseRequestExtensions.cs
andrueastman: @baywet I am hesitant to change this as the style used in the rest of the codebase to insert key values in the `MiddlewareOptions` collection is `typeof(AuthenticationHandlerOption).ToString()` which results to the string `AuthenticationHandlerOption` whereas `nameof(AuthenticationHandlerOption)` results in `Microsoft.Graph.AuthenticationHandlerOption`. A few examples [here](https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/e7ab26c8863b9ab0ca450ca95959d231205fad6a/src/Microsoft.Graph.Core/Extensions/BaseRequestExtensions.cs#L26), [here](https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/e7ab26c8863b9ab0ca450ca95959d231205fad6a/src/Microsoft.Graph.Core/Extensions/BaseRequestExtensions.cs#L51).

If I decide to change one, we could potentially cause a mismatch when reading from the `MiddlewareOptions` collection if we do not change them all correctly.

baywet: besides having to change it at other places, what would be the other negative impact potentially?

I'm worried about potential performance impacts where typeof does a complete reflection on the type but nameof only gets the fully qualified class name. Besides, I believe it should be typeof(Class).Name instead of ToString. In theory you never know when the ToString representation of the object will change (in practice it's unlikely they'll change it bit fear of breaking people's apps)

baywet: also, from what I can tell, nameof returns the class short name, not the fully qualified name https://dotnetfiddle.net/JIpBfi

andrueastman: > typeof(AuthenticationHandlerOption).ToString() which results to the string AuthenticationHandlerOption whereas nameof(AuthenticationHandlerOption) results in Microsoft.Graph.AuthenticationHandlerOption

Sorry about the confusion. This is actually the other way round. 

> I'm worried about potential performance impacts where typeof does a complete reflection on the type but nameof only gets the fully qualified class name.

Now I definitely have to fix this 😅. Since it will be in a separate PR, it will be much easier to track the specific change in case we miss something.

baywet: to provide hard evidence: nameof is just [syntax sugar and gets translated at compile time](https://www.erikschierboom.com/2015/12/31/csharp6-under-the-hood-nameof/), at runtime, it's just a constant string.
[typeof calls a bunch of methods under the hood at runtime](https://stackoverflow.com/questions/6417763/efficiency-of-cs-typeof-operator-or-whatever-its-representation-is-in-msil/6417977#6417977).
So it's always going to be faster to use nameof compared to typeof + Name.

andrueastman: @baywet Unfortunately, after attempting to do the refactoring, I don't think we can use `nameof` because of the function [here](https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/e7ab26c8863b9ab0ca450ca95959d231205fad6a/src/Microsoft.Graph.Core/Extensions/HttpRequestMessageExtensions.cs#L120). 😞 

Since we are using generics in that case to retrieve information from the `MiddlewareOptions` collection, `nameof` returns the string `T` rather than the key value that we would want. This is also covered in the restrictions section of your link earlier [here](https://www.erikschierboom.com/2015/12/31/csharp6-under-the-hood-nameof/#restrictions).

baywet: keep the typeof at that place, but read the Name property (shortname) instead of the ToString() return value, it should work :)

andrueastman: Thanks @baywet! This is now resolved via #275

andrueastman:
@baywet Any chance I can get a re-review? The review got reset after merging in #275

baywet (approved):

PR #275, opened 2021-07-15
Replace typeof(T).ToString() with nameof()

andrueastman:
This PR resolves feedback from https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/271

Currently we use `typeof(T).ToString()` to add/retrieve values from the `Middlewares` collection which introduces a performance penalty. This PR replaces those calls with `nameof` to alleviate that.

References
- https://www.erikschierboom.com/2015/12/31/csharp6-under-the-hood-nameof
- https://stackoverflow.com/questions/6417763/efficiency-of-cs-typeof-operator-or-whatever-its-representation-is-in-msil/6417977#6417977

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/275)


baywet (approved):
thanks for taking in the changes!

PR #296, opened 2021-08-03
Fix inconsistent serialization of interface properties

andrueastman:
The changes proposed in this PR involve

- Fixing the serialization of Interface properties by ensuring the JsonSerializerOptions intance is passed over to ensure consistent serialization with all types. 
- Fixing the DerivedTypeConverter to call the right  `JsonSerializer.Serialize` overload as the previous overload no longer supports a null `Type` parameter(See [here](https://docs.microsoft.com/en-us/dotnet/core/compatibility/serialization/5.0/jsonserializer-serialize-throws-argumentnullexception-for-null-type)).
- Adding/Updating tests to validate the above

Related issues.

- https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1084
- https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/issues/329

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/296)

### DIFF
--- a/src/Microsoft.Graph.Core/Serialization/DerivedTypeConverter.cs
+++ b/src/Microsoft.Graph.Core/Serialization/DerivedTypeConverter.cs
@@ -241,7 +241,17 @@ namespace Microsoft.Graph
                     foreach (var item in additionalData)
                     {
                         writer.WritePropertyName(item.Key);
-                        JsonSerializer.Serialize(writer, item.Value, item.Value.GetType(), options);
+                        // Check if value is null to choose the JsonSerializer.Serialize overload as System.Text.Json no longer supports 
+                        // the type parameter being null. 
+                        // Ref: https://docs.microsoft.com/en-us/dotnet/core/compatibility/serialization/5.0/jsonserializer-serialize-throws-argumentnullexception-for-null-type
+                        if (item.Value == null)
+                        {
+                            JsonSerializer.Serialize(writer, item.Value, options);
+                        }
+                        else
+                        {
+                            JsonSerializer.Serialize(writer, item.Value, item.Value.GetType(), options);
+                        }
                     }
                 }
                 else
--- a/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
+++ b/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
@@ -20,34 +20,10 @@
     <DelaySign>false</DelaySign>
     <AssemblyOriginatorKeyFile>35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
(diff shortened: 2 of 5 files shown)


src/Microsoft.Graph.Core/Serialization/DerivedTypeConverter.cs
```diff
@@ -241,7 +241,17 @@ public override void Write(Utf8JsonWriter writer, T value, JsonSerializerOptions
...
+                        // Check if value is null to choose the JsonSerializer.Serialize overload as System.Text.Json no longer supports 
+                        // the type parameter being null. 
+                        // Ref: https://docs.microsoft.com/en-us/dotnet/core/compatibility/serialization/5.0/jsonserializer-serialize-throws-argumentnullexception-for-null-type
+                        if (item.Value == null)
+                        {
+                            JsonSerializer.Serialize(writer, item.Value, options);
+                        }
+                        else
+                        {
+                            JsonSerializer.Serialize(writer, item.Value, item.Value.GetType(), options);
```
baywet, line 253:
```suggestion
                            JsonSerializer.Serialize(writer, item.Value, item.Value?.GetType(), options);
```

and remove the if/else

baywet (changes requested):

src/Microsoft.Graph.Core/Serialization/DerivedTypeConverter.cs
andrueastman: Hey @baywet,
The change is also because we can't do that anymore now as now the function now throws an `ArgumentNullException` for that parameter.
See the changelog at this link here. https://docs.microsoft.com/en-us/dotnet/core/compatibility/serialization/5.0/jsonserializer-serialize-throws-argumentnullexception-for-null-type

baywet: but isn't this passing null?
```CSharp
if (item.Value == null)
{
      JsonSerializer.Serialize(writer, item.Value, options);
}
```

andrueastman: Yes. The issue is the third parameter (the Type). So we use an overload that doesn't need the parameter.

baywet: aaah I see, sorry, morning here, I didn't register the difference in number of parameters somehow.

baywet (approved):

src/Microsoft.Graph.Core/Serialization/DerivedTypeConverter.cs
andrueastman: No worries 😄. Thanks for taking a look!

PR #299, opened 2021-08-09
Url decode nextLink url string.

andrueastman:
**Background**
https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/255 added support for the encoding of the query parameters in urls to fix issues arising from the use of special characters in QueryOptions. However, during paging, the `@odata.NextLink` property is returned from the api endpoint with space characters replaced with `+` which causes it to be encoded and result to errors. 

**Fix**
To fix this, the PR adds the new `NextLinkConverter` to the core library to fix the initialization of the nextLink property in collectionResponse objects by first decoding the relevant url(which will result in a url safe to encode like changing `+` to space representation).
The PR adds tests to validate that this is mitigated appropiately.

This PR should be followed by https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/570 which annotates the relevant property with the relevant converter so that we may encode the url safely.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/299)

### DIFF
--- a/src/Microsoft.Graph.Core/Serialization/NextLinkConverter.cs
+++ b/src/Microsoft.Graph.Core/Serialization/NextLinkConverter.cs
@@ -1,0 +1,47 @@
+﻿// ------------------------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+// ------------------------------------------------------------------------------
+
+namespace Microsoft.Graph
+{
+    using System;
+    using System.Net;
+    using System.Text.Json;
+    using System.Text.Json.Serialization;
+
+    public class NextLinkConverter : JsonConverter<string>
+    {
+        /// <summary>
+        /// Checks if the given object can be converted into a next link url.
+        /// </summary>
+        /// <param name="objectType">The object type.</param>
+        /// <returns>True if the object is of type Duration.</returns>
+        public override bool CanConvert(Type objectType)
+        {
+            return objectType == typeof(string);
+        }
+
+        /// <summary>
+        /// Deserialize the JSON data into a decoded nextLink url string.
+        /// </summary>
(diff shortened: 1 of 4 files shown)


src/Microsoft.Graph.Core/Serialization/NextLinkConverter.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        /// <summary>
+        /// Deserialize the JSON data into a decoded nextLink url string.
+        /// </summary>
+        /// <param name="reader">The <see cref="Utf8JsonReader"/> to read from.</param>
+        /// <param name="typeToConvert">The object type.</param>
+        /// <param name="options">The <see cref="JsonSerializerOptions"/> for conversion.</param>
+        /// <returns>A TimeOfDay object.</returns>
+        public override string Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
```
baywet, line 32: defensive programing

src/Microsoft.Graph.Core/Serialization/NextLinkConverter.cs
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        /// <summary>
+        /// Writes the JSON representation of the object.
+        /// </summary>
+        /// <param name="writer">The <see cref="Utf8JsonWriter"/> to write to.</param>
+        /// <param name="value">The nextLink url value.</param>
+        /// <param name="options">The calling serializer options</param>
+        public override void Write(Utf8JsonWriter writer, string value, JsonSerializerOptions options)
+        {
```
baywet, line 43: defensive programing

baywet (changes requested):

baywet (approved):

zengin (approved):

PR #337, opened 2021-11-22
Refactor BaseRequest/IResponseHandler/IHttpProvider

andrueastman:
This PR : - 

- Drops BaseRequest and its derivatives from the source to use `RequestInfomation`.
-  Refactors RequestBuilders to create `RequestInfomation` instances to use kiota core to send (batch and file upload).
- Drops `IHttpProvider` and its derivatives as to use `IRequestAdapter` from Kiota core
- Drops `IResponseHandler` to use the version from Kiota core and refactor its derivatives 
- Refactors relevant tests related to the above changes.

Notes :
- To make the PR easier to review by limiting its scope the `PageIterator` will be refactored in a separate PR

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/337)


src/Microsoft.Graph.Core/Requests/BatchRequestBuilder.cs
```diff
@@ -4,42 +4,67 @@
...
+        /// </summary>
+        /// <param name="batchRequestContent">The <see cref="BatchRequestContent"/> for the request</param>
+        /// <returns></returns>
+        public async Task<BatchResponseContent> PostAsync(BatchRequestContent batchRequestContent)
         {
-            return this.Request(null);
+            _ = batchRequestContent ?? throw new ArgumentNullException(nameof(batchRequestContent));
+            var requestInfo = await CreatePostRequestInformationAsync(batchRequestContent);
+            var responseHandler = new NativeResponseHandler();
+            await this.RequestAdapter.SendNoContentAsync(requestInfo, responseHandler);
```
baywet, line 48: out of curiosity, is there any specific reason why you're using the response handler instead of the provided deserialziation?

src/Microsoft.Graph.Core/Requests/Content/BatchRequestContent.cs
```diff
@@ -142,7 +146,8 @@ public string AddBatchRequestStep(IBaseRequest request)
                 });
 
             string requestId = Guid.NewGuid().ToString();
-            BatchRequestStep batchRequestStep = new BatchRequestStep(requestId, request.GetHttpRequestMessage());
+            var requestMessage = new HttpClientRequestAdapter(new AnonymousAuthenticationProvider()) {BaseUrl = baseClient.RequestAdapter.BaseUrl}.GetRequestMessageFromRequestInformation(requestInformation);
```
baywet, line 149: since you already have a request adapter on the client, why not use that instead of creating a new instance?

src/Microsoft.Graph.Core/Requests/Upload/UploadSliceRequestBuilder.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// <param name="totalSessionLength">Total session length. This MUST be consistent
+        /// across all slice.</param>
+        public UploadSliceRequestBuilder(
+            string sessionUrl,
+            IRequestAdapter requestAdapter,
+            long rangeBegin,
+            long rangeEnd,
+            long totalSessionLength)
+        {
+            this.RangeBegin = rangeBegin;
```
baywet, line 54: defensive programing

src/Microsoft.Graph.Core/Requests/AsyncMonitor.cs
```diff
@@ -42,61 +45,58 @@ public async Task<T> PollForOperationCompletionAsync(IProgress<AsyncOperationSta
         {
             while (!cancellationToken.IsCancellationRequested)
             {
-                using (var httpRequestMessage = new HttpRequestMessage(HttpMethod.Get, this.monitorUrl))
+                var requestInformation = new RequestInformation() { HttpMethod = Kiota.Abstractions.HttpMethod.GET , UrlTemplate = this.monitorUrl};
+                var nativeResponseHandler = new NativeResponseHandler();
+                await this.client.RequestAdapter.SendNoContentAsync(requestInformation, nativeResponseHandler).ConfigureAwait(false);
```
baywet, line 50: same comment about native response handling vs the serialization layer provided by kiota

baywet (changes requested):

src/Microsoft.Graph.Core/Requests/BatchRequestBuilder.cs
andrueastman: For the batch scenario, we typically return the native response object i.e. `HttpReponseMessage` and wrap it around the `BatchResponseContent` which would now do the job unwrapping the collection of responses received in the batch. Since in this scenario we need an instance of `HttpReponseMessage`, using the response handler is the way to go.

src/Microsoft.Graph.Core/Requests/AsyncMonitor.cs
andrueastman: I agree with you on this. I will action and resolve in an upcoming update.

src/Microsoft.Graph.Core/Requests/BatchRequestBuilder.cs
baywet: thanks for clarifying.
QQ: how do you find the handler pattern from a usability perspective?

baywet (approved):

src/Microsoft.Graph.Core/Requests/BatchRequestBuilder.cs
andrueastman: I think there's some great potential for it. 
Taking the example from the `NativeResponseHandler`, I'm realizing we could simply write a response handler that gives one back a `GraphResponse` object rather than the current method of having to generate parallel methods as we do [here](https://github.com/microsoftgraph/msgraph-sdk-dotnet/blob/6f133fc66b313b5355913e568a426bce25bfa8ad/src/Microsoft.Graph/Generated/requests/AadUserConversationMemberRequest.cs#L59) saving us a bit of work and then the user just passes in a parameter based on what/how the response is done. 

I am unsure how I feel though about having to do this with the `NativeResponseHandler` 
```cs
responseHandler.Value as HttpResponseMessage;
```
and maybe we could refactor the `NativeResponseHandler` to be `NativeResponseHandler<T>` so that the Value  property is already is strongly typed from the get go.

PR #339, opened 2021-11-29
Refactors the PageIterator to work with Kiota Core.

andrueastman:
This PR refactors the pageIterator to work with Kiota Core.

Changes include
- RequestConfigurator parameter configures instances of `RequestInformation`
- Next page requests are perfromed using RequestAdapter
- Page parameter is now expected to be an instance of `IParsable` from Kiota core libraries
- Updated tests to validate the changes.

Other changes
- Drop `Option` base class and derived instances(`HeaderOption` and `QueryOption`) as this is handled in Kiota Core
- Drop CollectionPage as collection responses contain collection that now are implementations of `IList`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/339)


src/Microsoft.Graph.Core/Tasks/PageIterator.cs
```diff
@@ -230,14 +233,50 @@ public async Task ResumeAsync()
...
+        /// </summary>
+        /// <param name="nextPageRequestInformation">The <see cref="RequestInformation"/> of the next page request</param>
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/> to use to cancel requests</param>
+        /// <returns></returns>
+        private async Task<IParsable> GetNextPageAsync(RequestInformation nextPageRequestInformation, CancellationToken cancellationToken)
+        {
+            // Use reflection to invoke the "requestAdapter.SendAsync" method so as to be able to be able to pass the generic type arguments
+            // neededed for handling the response.
+            var method = _client.RequestAdapter.GetType().GetMethod("SendAsync").MakeGenericMethod(_currentPage.GetType());
+            var sendTask = (Task)method.Invoke(_client.RequestAdapter, new object[] { nextPageRequestInformation, null, cancellationToken });
```
baywet, line 265: technically this is a task of the same type you get line 53. So by adding a second generic parameter to the class, with a constraint where U : IParsable, and reusing that generic type line 53 and here, you could avoid having to use reflection to get the result.

src/Microsoft.Graph.Core/Tasks/PageIterator.cs
```diff
@@ -230,14 +233,50 @@ public async Task ResumeAsync()
...
+            // return the result from the task.
+            return (IParsable)sendTask.GetType().GetProperty("Result").GetValue(sendTask);
+        }
+
+        /// <summary>
+        /// Helper method to extract the nextLink property from an <see cref="IParsable"/> instance.
+        /// </summary>
+        /// <param name="parsableCollection">The <see cref="IParsable"/> to extract the nextLink from</param>
+        /// <returns></returns>
+        private static string ExtractNextLinkFromParsable(IParsable parsableCollection)
```
baywet, line 276: make next link a parameter so you can reuse that method for delta link and affect delta link, I don't thing it's ever affected to anything other than string.empty now.

baywet (changes requested):

src/Microsoft.Graph.Core/Tasks/PageIterator.cs
andrueastman: Thats true.
I was mainly at a bit of a crossroads between doing this versus the reflection route as this means that the developer's experience is altered in the way that creating a Page Iterator changes from this

```cs
            eventPageIterator = PageIterator<Event>.CreatePageIterator(baseClient, originalPage, (e) =>
            {
                events.Add(e);
                return true;
            });
```

to this (where the developer has the extra work to specify the collection type and I couldn't tell if it was worth it from their perspective, so I was trying to save the hassle). However, from our perspective I really appreciate the type safety.

```cs
eventPageIterator = PageIterator<Event, EventsResponse>.CreatePageIterator(baseClient, originalPage, (e) =>
            {
                events.Add(e);
                return true;
            });
```

This is now resolved in the latest update.

src/Microsoft.Graph.Core/Tasks/PageIterator.cs
andrueastman: resolved in latest update.

baywet (approved):

PR #350, opened 2022-01-03
[Kiota] LargeFileUpload task refactoring 

andrueastman:
This PR refactors the LargeFileUploadTask to be work with the kiota core libraries in conjuction with refactoring made in previous PRs. 

Changes include
- Refactoring the  LargeFileUploadTask constructor to take an IParsable that is verified to be an uploadSession
- Refactoring relevant models
- Refactoring tests

**Notes**
- There is an open issue with regards to creating an uploadSession with the generated request builders [here](https://github.com/microsoft/kiota/issues/990). 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/350)

### DIFF
--- a/src/Microsoft.Graph.Core/Models/UploadSession.cs
+++ b/src/Microsoft.Graph.Core/Models/UploadSession.cs
@@ -4,14 +4,24 @@
 
 namespace Microsoft.Graph.Core.Models
 {
+    using Microsoft.Kiota.Abstractions.Serialization;
     using System;
     using System.Collections.Generic;
+    using System.Linq;
 
     /// <summary>
     /// Concrete implementation of the IUploadSession interface
     /// </summary>
-    internal class UploadSession : IUploadSession
+    internal class UploadSession : IUploadSession,IParsable
     {
+        /// <summary>
+        /// Instantiates a new uploadSession and sets the default values.
+        /// </summary>
+        public UploadSession()
+        {
+            AdditionalData = new Dictionary<string, object>();
+        }
+
         /// <summary>
         /// Expiration date of the upload session
         /// </summary>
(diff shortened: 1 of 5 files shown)


src/Microsoft.Graph.Core/Models/UploadSession.cs
```diff
@@ -26,5 +36,36 @@ internal class UploadSession : IUploadSession
         /// The URL for upload
         /// </summary>
         public string UploadUrl { get; set; }
+
+        /// <summary>
+        /// Stores additional data not described in the OpenAPI description found when deserializing. Can be used for serialization as well.
+        /// </summary>
+        public IDictionary<string, object> AdditionalData { get; set; }
```
baywet, line 43:
```suggestion
        public IDictionary<string, object> AdditionalData { get; set; } = new Dictionary<string, object>();
```

and drop the constructor

src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
```diff
@@ -33,24 +35,48 @@ public class LargeFileUploadTask<T>
         /// <param name="maxSliceSize">Max size of each slice to be uploaded. Multiple of 320 KiB (320 * 1024) is required.</param>
         /// <param name="baseClient"><see cref="BaseClient"/> to use for making upload requests. The client should not set Auth headers as upload urls do not need them.
         /// If less than 0, default value of 5 MiB is used. .</param>
-        public LargeFileUploadTask(IUploadSession uploadSession, Stream uploadStream,  int maxSliceSize = -1, BaseClient baseClient = null)
+        public LargeFileUploadTask(IParsable uploadSession, Stream uploadStream,  int maxSliceSize = -1, BaseClient baseClient = null)
```
baywet, line 38: push the IParsable interface on IUploadSession, revert this change in the ctor, this way you don't have to juggle down below to get the actual session object.

baywet (changes requested):

src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
andrueastman: I agree that would be better. I will need to run some tests first then will action and resolve.

andrueastman: Resolved via https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/350/commits/a649908819c3c2e1f93e878bfbeb2236886b1b85

src/Microsoft.Graph.Core/Models/UploadSession.cs
```diff
@@ -4,8 +4,10 @@
 
 namespace Microsoft.Graph.Core.Models
 {
+    using Microsoft.Kiota.Abstractions.Serialization;
```
baywet, line 7: I'm curious to what's the plan for this "handcrafted" upload session vs the code-gen ones?
The customer will have access to something like `client.Me.Drive.Items["id"].Value.CreateUploadSession.PostAsync()`, which will return an upload session from the code-gen, looking a lot like this one, but not implementing the interface.
Now the large file upload task itself takes an IUploadSession, which is implemented by this class. How do you plan people going from what the fluent API returned to this?

baywet (commented):
thanks for taking the changes, I do have an additional question from a usability perspective.

src/Microsoft.Graph.Core/Models/UploadSession.cs
andrueastman: This would ideally be achieved through extending the model class in the service library. In my head we could achieve this using partial classes as we already do currently. 
The alternative would be to use your suggestion of a config file, which would simply enrich matching classes with relevant code element properties (in this case implementing the interface).

baywet: partial classes are a very dotnet approach to this issue, I'd like to address this aspect in a similar way across languages if we can to ease up the maintenance burden on language owner. I'll go ahead and capture what we said in the issue you linked. For now we can keep this duplicate class, I'll go ahead and approve this PR, knowing we'll have to finalize the work once the extensibility story is tackled in kiota.

baywet (approved):

PR #356, opened 2022-01-17
Odata query handler

andrueastman:
This PR adds an OData query handler to the middleware pipeline as the generated query parameters from Kiota lack the `$` sign when requests are sent out.

As much as this should not be a problem in OData 4.0 which no longer requires the `$` prefix some workloads on graph still do not support this and this middleware alleviates the errors that may arise from a missing `$` prefix by reinserting based on whether the query parameter is an OData query parameter.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/356)

### DIFF
--- a/src/Microsoft.Graph.Core/Requests/Middleware/OdataQueryHandler.cs
+++ b/src/Microsoft.Graph.Core/Requests/Middleware/OdataQueryHandler.cs
@@ -1,0 +1,61 @@
+﻿// ------------------------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+// ------------------------------------------------------------------------------
+
+namespace Microsoft.Graph
+{
+    using Microsoft.Kiota.Http.HttpClientLibrary.Extensions;
+    using System;
+    using System.Net.Http;
+    using System.Text.RegularExpressions;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// A <see cref="DelegatingHandler"/> implementation that handles compression.
+    /// </summary>
+    public class OdataQueryHandler : DelegatingHandler
+    {
+        private readonly ODataQueryHandlerOption odataQueryHandlerOption;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Graph.Core/Requests/Middleware/OdataQueryHandler.cs
```diff
@@ -0,0 +1,61 @@
...
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// A <see cref="DelegatingHandler"/> implementation that handles compression.
+    /// </summary>
+    public class OdataQueryHandler : DelegatingHandler
+    {
+        private readonly ODataQueryHandlerOption odataQueryHandlerOption;
+        private readonly Regex odataQueryRegex = new Regex("(?i)([^$])(count|deltatoken|expand|filter|format|orderby|search|select|skip|skiptoken|top)=",RegexOptions.Compiled);
```
baywet, line 20: nit: dotnet developers usually use regexOptions.ignoreCase instead of the prag match syntax

src/Microsoft.Graph.Core/Requests/Middleware/OdataQueryHandler.cs
```diff
@@ -0,0 +1,61 @@
...
+                var queryString = string.Empty;
+                if (httpRequest.RequestUri.Query != null && httpRequest.RequestUri.Query.Length > 1)
+                    // We insert and remove the ? sign so we can make no dollar mandatory and avoid adding a second dollar when already here
+                    queryString = odataQueryRegex.Replace("?" + httpRequest.RequestUri.Query, "$1$$$2=")[1..];
+
+                // replace the uri with the new query options
+                httpRequest.RequestUri = new UriBuilder(httpRequest.RequestUri) { Query = queryString }.Uri;
+            }
+
+            HttpResponseMessage response = await base.SendAsync(httpRequest, cancellationToken);
```
baywet, line 56: you can return the response directly. and remove await/async

baywet (changes requested):
just a few minor comments.

baywet (approved):

PR #359, opened 2022-01-21
Setup BaseRequestAdapter

andrueastman:
This PR :-

- Adds a `BaseRequestAdapter` to enable passing of custom options to setup graphClient and easy configuration/creation of a `IRequestAdapter` instance that works well with graph.
- Configures the telemetry handler to send the telemetry according to the telemetry spec and allows headers to be configured with `GraphClientOptions` for dependencies like graph cli
- Changes the `NextExpectedRanges` property type in `IUploadSession`from `IEnumerable` to `List` as Kiota generates collection properties with the `List` type.
- Relevant test refactoring.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/359)


src/Microsoft.Graph.Core/Requests/GraphClientFactory.cs
```diff
@@ -262,5 +263,44 @@ private static Uri DetermineBaseAddress(string nationalCloud, string version)
...
 
         }
+
+        internal static TelemetryHandler GetConfiguredTelemetryHandler(GraphClientOptions graphClientOptions)
+        {
+            // Build the service library string from the options
+            var serviceLibraryString = string.Empty;
+            if (!string.IsNullOrEmpty(graphClientOptions?.GraphServiceLibraryClientVersion))
+            {
+                serviceLibraryString = graphClientOptions?.GraphProductPrefix ?? "graph-dotnet";
```
baywet, line 273: what about moving the header build-up logic to the handler itself?

baywet (commented):
Thanks for the quick turnaround on this one. One suggestion.

src/Microsoft.Graph.Core/Requests/GraphClientFactory.cs
andrueastman: Thanks for pointing this out. This was inconsistent with other SDKs(e.g Go and the spec here), and should be resolved in the latest update.

src/Microsoft.Graph.Core/Requests/BaseClient.cs
```diff
@@ -70,7 +69,7 @@ IAuthenticationProvider authenticationProvider
         /// <param name="httpClient">The customized <see cref="HttpClient"/> to be used for making requests</param>
         public BaseClient(
             string baseUrl,
-            HttpClient httpClient):this(new HttpClientRequestAdapter(new AnonymousAuthenticationProvider(), httpClient: httpClient) { BaseUrl = baseUrl })
+            HttpClient httpClient):this(new BaseGraphRequestAdapter(new AnonymousAuthenticationProvider(), httpClient: httpClient) { BaseUrl = baseUrl })
```
baywet, line 72: why are we defaulting to an anonymous authentication provider here instead of requiring the authentication provider? especially since we don't have an auth handler anymore.

baywet (changes requested):
thanks for taking in the changes, one last thing.

src/Microsoft.Graph.Core/Requests/BaseClient.cs
andrueastman: This constructor has a couple of jobs
- Allow simple creating of a client with no authentication provider in the event the user wishes to have custom auth in the http middleware
- Create clients for scenarios like large file upload where there is no need for auth. 

In the event one has an authProvider/tokencredential, they will use the other two constrcutors above.

baywet: ok, thanks, we can always review the usefulness of the base client class later since there is no way for the generated graph service client to inherit from it today. resolving.

baywet (approved):

PR #392, opened 2022-03-18
Adds IBaseClient for extensibility

andrueastman:
This includes the following changes;

- Adds an IBaseClient for extensibility with the service library
- Adds back the `ExtractSessionFromParsable` method for handling upload sessions for the LargeFileUploadTask to accept an IParsable for now. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/392)

### DIFF
--- a/src/Microsoft.Graph.Core/Requests/BaseClient.cs
+++ b/src/Microsoft.Graph.Core/Requests/BaseClient.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Graph
     /// <summary>
     /// A default client implementation.
     /// </summary>
-    public class BaseClient
+    public class BaseClient: IBaseClient
     {
         /// <summary>
         /// Constructs a new <see cref="BaseClient"/> for use by derived instances.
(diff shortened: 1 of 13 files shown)


src/Microsoft.Graph.Core/Requests/BaseClient.cs
```diff
@@ -17,7 +17,7 @@ namespace Microsoft.Graph
     /// <summary>
     /// A default client implementation.
     /// </summary>
-    public class BaseClient
```
baywet, line 20: I think this class should be moved to the service lib, renamed GraphServiceClient and inherit from the GraphBaseServiceClient (generated by Kiota) for the full loop to be closed on this. Otherwise you'll end up needing to inherit from 2 classes in the service lib, not possible in dotnet.

baywet (changes requested):
Thanks for taking a look at this. I think there is a minor design change to make here.

src/Microsoft.Graph.Core/Requests/BaseClient.cs
andrueastman: Actually, this class is only meant to be used internally in the core lib really. For testing and when a concrete implementation of the interface is needed. 
I've made it internal to stay consistent with the plan if that makes sense.

baywet: thanks for the clarification!

baywet (approved):

PR #408, opened 2022-04-12
Drop odataQueryHandler

### LINKED ISSUE #406: Remove the ODataQueryParametersHandler
For the preview v1 version, once these are merged and the references are updated:
- https://github.com/microsoft/kiota/pull/1498
- https://github.com/microsoft/kiota-http-dotnet/pull/12
- https://github.com/microsoft/kiota-http-dotnet/pull/12

(and once a generation had been triggered)

andrueastman:
This PR closes #406 

It drops the `ODataQueryParametersHandler` in favour of the `ParametersNameDecodingHandler` available from Kiota core.

Tests have also been refactored for this changes. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/408)


src/Microsoft.Graph.Core/Requests/GraphClientFactory.cs
```diff
@@ -112,7 +112,7 @@ public static IList<DelegatingHandler> CreateDefaultHandlers(GraphClientOptions
         {
             return new List<DelegatingHandler> {
                 new GraphTelemetryHandler(graphClientOptions),
```
baywet, line 114: is there any specific reason here why we don't call the kiota factory to get the default handlers for kiota and just add the telemetry on top of that?

baywet (changes requested):
doesn't the csproj need to be updated as well with the new references?

src/Microsoft.Graph.Core/Requests/GraphClientFactory.cs
andrueastman: Thanks for calling this out. 
This was a means to control the handler order due to do some remaining work in https://github.com/microsoft/kiota/issues/598.

I've now incorporated proposals made from https://github.com/microsoftgraph/msgraph-sdk-design/pull/75 and added the call to the kiota factory.

baywet (approved):

andrueastman:
Thanks @baywet

PR #474, opened 2022-08-05
Remove restriction on 320KB multiple restriction for the LargeFileUploadTask

andrueastman:
This PR closes https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1436

It removes the Argument Exception thrown when the LargeFileUploadTask is initialized with a slice size that is not a multiple of 320KB to allow for other scenarios like PrintDocument. 
This makes the LargeFileUploadTask more agnostic to the endpoints being uploaded to and leaves the failure of the task to the exception thrown by the API in the event of an incorrect slice size.

Tests have been updated to capture this as well. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/474)

### DIFF
--- a/src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
+++ b/src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
@@ -16,7 +16,6 @@ namespace Microsoft.Graph
     public class LargeFileUploadTask<T>
     {
         private const int DefaultMaxSliceSize = 5 * 1024 * 1024;
-        private const int RequiredSliceSizeIncrement = 320 * 1024;
         private IUploadSession Session { get; set; }
         private readonly IBaseClient _client;
         private readonly Stream _uploadStream;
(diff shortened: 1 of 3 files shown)


src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
```diff
@@ -30,7 +29,7 @@ public class LargeFileUploadTask<T>
         /// </summary>
         /// <param name="uploadSession">Session information of type <see cref="IUploadSession"/>></param>
         /// <param name="uploadStream">Readable, seekable stream to be uploaded. Length of session is determined via uploadStream.Length</param>
-        /// <param name="maxSliceSize">Max size of each slice to be uploaded. Multiple of 320 KiB (320 * 1024) is required.</param>
+        /// <param name="maxSliceSize">Max size(in bytes) of each slice to be uploaded. Defaults to 5MB</param>
```
baywet, line 32:
```suggestion
        /// <param name="maxSliceSize">Max size(in bytes) of each slice to be uploaded. Defaults to 5MB. When uploading to OneDrive or SharePoint, this value needs to be a multiple of 320 KiB (327,680 bytes).</param>
```

baywet (dismissed):

src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
andrueastman: Thanks!

baywet (approved):

PR #484, opened 2022-08-22
Add `Object` postfix to type name in `DerivedTypeConverter`

KnapSac:
Fixes microsoftgraph/msgraph-sdk-dotnet#1449.

For Graph resources which end with `Request`, the model generator adds
the postfix `Object`. Since we use the ODataType to construct an
instance of the generated class in `DerivedTypeConverter`, we actually
attempt to create instances of the wrong type. To fix this, we also add
the postfix to the type name in `DerivedTypeConverter`, before
attempting to create instances of the type.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/484)

### DIFF
--- a/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
+++ b/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
@@ -21,10 +21,10 @@
     <DelaySign>false</DelaySign>
     <AssemblyOriginatorKeyFile>35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
-    <VersionPrefix>2.0.11</VersionPrefix>
+    <VersionPrefix>2.0.12</VersionPrefix>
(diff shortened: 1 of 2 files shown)


andrueastman (commented):
Thanks for this @KnapSac 

Any chance you can also push an update to the csproj file to bump the version and add a note for the changes made?

https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/239abe6054ddb483ebb4888b005c2ea879a4a2a9/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj#L27

andrueastman (approved):
Thanks for this!

KnapSac:
@andrueastman do you have any idea when we can expect a new release? I need this fix, and I would rather not mess around with referencing a dev build if I don't have to.

andrueastman:
Hey @KnapSac, 
The changes should now be now available through the latest [nuget update](https://www.nuget.org/packages/Microsoft.Graph.Core/2.0.12).

Issue #227, opened 2021-03-25
Add ability to build the project with dotnet CLI


zengin:
Currently the project can only be built by msbuild. `dotnet build` hits the following error:
```
C:\github\msgraph-sdk-dotnet-core\src\Microsoft.Graph.Core\Microsoft.Graph.Core.csproj :
  error MSB4057: The target "_GetRestoreSettingsPerFramework" does not exist in the project.
```

This might be relevant: https://github.com/novotnyllc/MSBuildSdkExtras/issues/85 
[AB#8736](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8736)

maisarissi:
Hi @zengin . This bug is due to .Net SDK TFMs limitations:  https://github.com/dotnet/sdk/issues/491

andrueastman:
Closed via https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/486

Issue #39, opened 2019-09-26
Can we get a .sln file that doesn't require the Xamarin SDK installed?


darrelmiller:
It's not very friendly to require that the Android SDK is installed to open the solution.  Also, we should provide installation instructions in the readme as to what Android related SDKs do need to be installed. 
[AB#7198](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7198)

andrueastman:
Closed via https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/486

PR #486, opened 2022-08-24
MAUI support

### LINKED ISSUE #373: Passing handlers to GraphClientFactory.Create() creates a HttpClient that disposes the handlers

### Expected behavior
Documentation of [GraphClientFactory.Create()](https://docs.microsoft.com/en-us/dotnet/api/microsoft.graph.graphclientfactory.create?view=graph-core-dotnet#microsoft-graph-graphclientfactory-create(system-collections-generic-ienumerable((system-net-http-delegatinghandler))-system-string-system-string-system-net-iwebproxy-system-net-http-httpmessagehandler)) method does not describe disposal of passed handlers, eg the collection of `DelegatingHandler` and optional `finalHandler`.

IMO, passing disposable objects to another object does not imply that the receiving object can take ownership of those disposable objects. However changing the behavior is breaking, and adding an additional optional argument to control dispose behavior may not be desirable. So maybe this is basically just a documentation issue.

### Actual behavior
Currently the method creates a `HttpClient` that disposes its `HttpMessageHandler` and thus the collection of `DelegatingHandler` and optional `finalHandler` as the `HttpClient` is created without specifying the `bool disposeHandler` parameter.

A test that shows finalHandler is disposed:

```csharp
[Fact]
public void CreateClient_WithFinalHandlerDisposesTheFinalHandler()
{
    var finalHandler = new MockHttpHandler();
    using (HttpClient client = GraphClientFactory.Create(handlers: GraphClientFactory.CreateDefaultHandlers(testAuthenticationProvider.Object), finalHandler: finalHandler))
    {
        Assert.NotNull(client);
    }

    Assert.True(finalHandler.Disposed);
}

private class MockHttpHandler : HttpMessageHandler
{
    public bool Disposed;

    protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
    {
        throw new NotImplementedException();
    }

    protected override void Dispose(bool disposing)
    {
        Disposed = true;
        base.Dispose(disposing);
    }
}
```

andrueastman:
Changes include: 
- [Breaking] Replacing Xamarin build targets with MAUI build targets. Xamarin users can use the MAUI build targets for ios android and macos. Closes https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/issues/413
- Enable the use of building the project with .NET CLI as https://github.com/dotnet/sdk/issues/491 is resolved using MAUI targets to close https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/issues/227
- Removes the `Microsoft.Graph.Core.XamarinAndroid.Test` project as the `net6.0` target is sufficient to close https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/issues/39
- Updates tests and test project to `net6.0` target
- Updates the GraphClientFactory.Create method to allow the HttpClient to use the extra constructor that prevents handler disposal to close #373 
- Updates CI/CD pipelines to use dotnet CLI commands rather than `msbuild` 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/486)


src/Microsoft.Graph.Core/Requests/Content/BatchResponseContent.cs
```diff
@@ -177,7 +177,7 @@ private HttpResponseMessage GetResponseMessageFromJObject(JsonElement jResponseI
         /// <returns>A batch response content as <see cref="JsonDocument"/>.</returns>
         private async Task<JsonDocument> GetBatchResponseContentAsync()
         {
-            if (this.batchResponseMessage.Content == null)
+            if (this.batchResponseMessage.Content == null || this.batchResponseMessage.Content.Headers.ContentLength == 0 )
```
baywet, line 180:
```suggestion
            if (this.batchResponseMessage.Content == null || !this.batchResponseMessage.Content.Headers.Any())
```

PR #514, opened 2022-09-30
PageIterator/LargeFileUploadTask enhancements

andrueastman:
This PR adds contructors/factories for PageIterator/LargeFileUploadTask that accept IRequestAdapter to also better serve scenarios where sdks are self-generated and thus do not have defined implementation for IBaseClient as the generated service libraries do.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/514)

### DIFF
--- a/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
+++ b/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
@@ -21,11 +21,9 @@
     <AssemblyOriginatorKeyFile>35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
     <VersionPrefix>3.0.0</VersionPrefix>
-    <VersionSuffix>preview.14</VersionSuffix>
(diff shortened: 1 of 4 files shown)


baywet (changes requested):
Following up on the conversation with @ddyett I think the point was to be able to initialize those with either the client or the request adapter without the need for a drop-in client (in case of self-serve).

On top of the changes you've just made, I think we should add overload which accept IBaseClient instead, and just call the request adapter overload with client.GetRequestAdapter()

What do you think?

andrueastman:
Hey @baywet,

Any chance you could clarify what I missed, I figure that is what I already did with adding the "overloads" like this one. I realise now the PR diff may not show that too clearly. 

https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/aeb7266c7898895f9f911f2ca6d3f0ca67eed239/src/Microsoft.Graph.Core/Tasks/PageIterator.cs#L115

baywet (approved):
Ah sorry, ignore my confuse and decaffeinated self, thanks for the clarification

PR #547, opened 2022-12-06
Omit CompressionHandler in Macatalyst 

andrueastman:
This PR includes MACCATALYST preprocessor directive to omit CompressionHandler that is already handler by the NSUrlSessionHandler

It also aligns other directives to the directives used in the codebase. 

Closes https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1551

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/547)


baywet (commented):
Will the conditional build directives work though? The application can be built in Windows and then distributed to a Mac machine.

andrueastman:
@baywet, They will, as the `csproj` defines the respective targets, and results in producing artifacts for each of the targets which will be used depending on the target system the app developer chooses. 

E.g building build an app with `macatalyst` as a target will use the appropriate artifacts.

baywet (approved):
thanks for clarifying, I had missed that part of the directives configuration, I thought they were the default from the framework

PR #607, opened 2023-02-28
Error handling cleanup

andrueastman:
Cleans up error handling code to avoid code duplication of the Error model.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/607)


src/Microsoft.Graph.Core/Exceptions/ClientException.cs
```diff
@@ -9,14 +9,14 @@ namespace Microsoft.Graph
     /// <summary>
     /// Graph client exception.
     /// </summary>
-    public class ClientException : ServiceException
+    public class ClientException : Exception
```
baywet, line 12: should we have this one also inherit api exception so consumers can implement generic try catch clauses?

baywet (commented):
There's one thing I'm not clear on: in case of batching, where is the error mapping from the service generated error types being passed?

andrueastman:
> There's one thing I'm not clear on: in case of batching, where is the error mapping from the service generated error types being passed?

This was missing but resolved in the latest update.

src/Microsoft.Graph.Core/Requests/Content/BatchResponseContent.cs
```diff
@@ -86,17 +82,18 @@ public async Task<HttpResponseMessage> GetResponseByIdAsync(string requestId)
         /// Gets a batch response as a requested type for the specified batch request id.
         /// </summary>
         /// <param name="requestId">A batch request id.</param>
-        /// <param name="responseHandler">ResponseHandler to use for the resonse</param>
+        /// <param name="errorMappings">A batch request id.</param>
+        /// <param name="responseHandler">ResponseHandler to use for the response</param>
         /// <returns>A deserialized object of type T<see cref="HttpResponseMessage"/>.</returns>
-        public async Task<T> GetResponseByIdAsync<T>(string requestId, IResponseHandler responseHandler = null) where T : IParsable, new()
+        public async Task<T> GetResponseByIdAsync<T>(string requestId, Dictionary<string, ParsableFactory<IParsable>> errorMappings = null, IResponseHandler responseHandler = null) where T : IParsable, new()
```
baywet, line 88: so here we're expecting the caller to pass that mapping to us if I'm not mistaking? 
What about pushing that argument to the constructor, storing the value as a private field.
And then have derived classes in the service lib instead?

andrueastman: https://github.com/microsoftgraph/msgraph-sdk-dotnet/blob/c065dd3405174a6aad94e85859b7abb1b673baa5/src/Microsoft.Graph/GraphServiceClient.cs#L92

andrueastman: Thanks for clearing this up. 
The latest update reverts this and makes it a constructor variable. https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/607/commits/837b328b02764447041978953f3d94f426c92272
The service library can override this in its request builder here https://github.com/microsoftgraph/msgraph-sdk-dotnet/blob/c065dd3405174a6aad94e85859b7abb1b673baa5/src/Microsoft.Graph/GraphServiceClient.cs#L92

src/Microsoft.Graph.Core/Requests/Content/BatchResponseContent.cs
```diff
@@ -22,18 +22,17 @@ public class BatchResponseContent
...
+        /// <param name="errorMappings">A dictionary of error mappings to handle failed responses.</param>
+        public BatchResponseContent(HttpResponseMessage httpResponseMessage, Dictionary<string, ParsableFactory<IParsable>> errorMappings = null)
         {
-            this.batchResponseMessage = httpResponseMessage ?? throw new ClientException(new Error
-            {
-                Code = ErrorConstants.Codes.InvalidArgument,
-                Message = string.Format(ErrorConstants.Messages.NullParameter, nameof(httpResponseMessage))
-            });
+            this.batchResponseMessage = httpResponseMessage ?? throw new ArgumentNullException(nameof(httpResponseMessage));
+            this.apiErrorMappings = errorMappings;
```
baywet, line 35:
```suggestion
            this.apiErrorMappings = errorMappings ?? new();
```

baywet (changes requested):

baywet (approved):
Thanks for making these changes

src/Microsoft.Graph.Core/Requests/Content/BatchResponseContent.cs
andrueastman: 👍🏼

Issue #554, opened 2022-12-15
LargeFileUploadTask with a zero length stream throws TaskCanceledException "Upload failed too many times"


redback790:
Please provide the following (and please check them off the list with [x]) before submitting this issue:
- [x] Expected behavior. Please provide **links to the specific [Microsoft Graph documentation](https://developer.microsoft.com/en-us/graph/docs/concepts/overview)** you used to determine the expected behavior. 
- [x] Actual behavior. Provide error codes, stack information, and a [Fiddler](http://www.telerik.com/fiddler) capture of the request and response (please remove personally identifiable information before posting).
- [x] Steps to reproduce the behavior. Include your code, IDE versions, client library versions, and any other information that might be helpful to understand your scenario.

### Expected behavior

Using an instance of LargeFileUploadTask.cs with zero length Stream should either create a zero-sized file or throw ArgumentException.

For a small sized stream, e.g. 4 bytes, it works fine and a new small file is created at the requested SharePoint path.

### Actual behavior

There is an unhelpful exception thrown: 
```
Unhandled exception. System.Threading.Tasks.TaskCanceledException: Upload failed too many times. See InnerException for list of exceptions that occured.
 ---> System.AggregateException: One or more errors occurred.
   --- End of inner exception stack trace ---
   at Microsoft.Graph.LargeFileUploadTask`1.UploadAsync(IProgress`1 progress, Int32 maxTries)
```

### Steps to reproduce the behavior

Microsoft.Graph 4.47.0

Example code below:
```
string localFilePath = Path.Join(Path.GetTempPath(), "testzerosizedfile.tmp");
using (File.Create(localFilePath, 0, FileOptions.None)) {} // create zero-sized file
using FileStream fileStream = new FileStream(localFilePath, FileMode.Open);
IDriveItemRequestBuilder requestBuilder = graphClient.Drives[driveId].Root.ItemWithPath("testzerosizedfile.tmp");
UploadSession uploadSession = await requestBuilder.CreateUploadSession().Request().PostAsync();
LargeFileUploadTask<DriveItem> fileUploadTask = new LargeFileUploadTask<DriveItem>(uploadSession, content, maxSliceSize);
UploadResult<DriveItem> uploadResult = await fileUploadTask.UploadAsync();
```

a-bondarchuk-kdd:
[...]

PR #620, opened 2023-03-10
Fixes deltaLink parsing in PageIterator.

### LINKED ISSUE #619: PageIterator doesn't give the delta link after completing the iteration
`PageIterator` does not return the `@odata.nextLink` property, making it unusable for delta queries.

### Expected behavior

If you use the `PageIterator` class to iterate through a delta response, the `Deltalink` property should contain the value of the `@odata.deltaLink` property returned by the API. This value should be usable to create a new `DeltaRequestBuilder`.

### Actual behavior

The `Deltalink` property is empty (`""`). The value is there in the iterator's private `_currentPage.OdataDeltaLink` property:

image

`Microsoft.Graph` version 5.1.0

```csharp
// Local list of mail folders
var localMailFolders = new List<MailFolder>();

try
{
    // Get the first page of folders
    var mailFoldersDelta = await graphClient.Me
        .MailFolders
        .Delta
        .GetAsync();

    while (mailFoldersDelta != null)
    {
        var iterator = PageIterator<MailFolder, DeltaResponse>
            .CreatePageIterator(graphClient, mailFoldersDelta, async folder =>
            {
                localMailFolders.Add(folder);
                return true;
            });

        await iterator.IterateAsync();

        Console.WriteLine($"Sync complete. Delta link: {iterator.Deltalink}");
        Console.WriteLine($"Waiting for {pollInterval} seconds...");
        await Task.Delay(pollInterval * 1000);

        // Following line throws ArgumentNullException
        var deltaRequest = new DeltaRequestBuilder(iterator.Deltalink, graphClient.RequestAdapter);
        mailFoldersDelta = await deltaRequest.GetAsync();
    }
}
catch (Exception ex)
{
    Console.WriteLine($"Error: {ex.Message}");
    if (ex is ODataError odataErr)
    {
        Console.WriteLine(odataErr.Message);
    }
}
```

andrueastman:
Closes #619 
Closes https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/issues/554

It fixes parsing of the deltaLink in the pageIterator as the deltaLink will not necessarily live in the additionalData but as a strongly typed property of the `BaseDeltaFunctionResponse` class.

Test models have been updated to capture this. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/620)

### DIFF
--- a/src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
+++ b/src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
@@ -43,6 +43,10 @@ namespace Microsoft.Graph
             {
                 throw new ArgumentException("Must provide stream that can read and seek");
             }
+            if (uploadStream.Length < 1)
+            {
+                throw new ArgumentException("Must a stream that is not empty");
+            }
             this.Session = ExtractSessionFromParsable(uploadSession);
             this._requestAdapter = requestAdapter ?? this.InitializeAdapter(Session.UploadUrl);
             this._uploadStream = uploadStream;
--- a/src/Microsoft.Graph.Core/Tasks/PageIterator.cs
+++ b/src/Microsoft.Graph.Core/Tasks/PageIterator.cs
@@ -354,11 +354,11 @@ namespace Microsoft.Graph
         private static string ExtractNextLinkFromParsable(TCollectionPage parsableCollection, string nextLinkPropertyName = "OdataNextLink")
         {
             var nextLinkProperty = parsableCollection.GetType().GetProperty(nextLinkPropertyName);
-            if (nextLinkProperty != null)
+            if (nextLinkProperty != null && 
+                nextLinkProperty.GetValue(parsableCollection, null) is string nextLinkString  
+                && !string.IsNullOrEmpty(nextLinkString))
             {
-                var nextLinkString = nextLinkProperty.GetValue(parsableCollection, null) as string ?? string.Empty;
-                if (!string.IsNullOrEmpty(nextLinkString))
-                    return nextLinkString;
+                return nextLinkString;
             }
             
             // the next link property may not be defined in the response schema so we also check its presence in the additional data bag
--- a/src/Microsoft.Graph.Core/Tasks/PageIterator.cs
+++ b/src/Microsoft.Graph.Core/Tasks/PageIterator.cs
@@ -183,9 +183,19 @@ namespace Microsoft.Graph
             }
 
             // There are no pages CURRENTLY ready to be paged. Attempt to call delta query later.
-            else if (_currentPage.AdditionalData != null && _currentPage.AdditionalData.TryGetValue(CoreConstants.OdataInstanceAnnotations.DeltaLink, out object deltalink))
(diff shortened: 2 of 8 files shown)


src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
```diff
@@ -43,6 +43,10 @@ public LargeFileUploadTask(IParsable uploadSession, Stream uploadStream, int max
             {
                 throw new ArgumentException("Must provide stream that can read and seek");
             }
+            if (uploadStream.Length < 1)
```
baywet, line 46: doesn't this put us at risk of getting a not supported exception if they are streaming from another http service for example without buffering in memory?
https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.length?view=net-7.0#exceptions

src/Microsoft.Graph.Core/Tasks/PageIterator.cs
```diff
@@ -346,8 +356,11 @@ private static string ExtractNextLinkFromParsable(TCollectionPage parsableCollec
             var nextLinkProperty = parsableCollection.GetType().GetProperty(nextLinkPropertyName);
             if (nextLinkProperty != null)
             {
-                return nextLinkProperty.GetValue(parsableCollection, null) as string ?? string.Empty;
+                var nextLinkString = nextLinkProperty.GetValue(parsableCollection, null) as string ?? string.Empty;
+                if (!string.IsNullOrEmpty(nextLinkString))
+                    return nextLinkString;
```
baywet, line 361:
```suggestion
                return nextLinkProperty.GetValue(parsableCollection, null) is string nextLinkString  && !string.IsNullOrEmpty(nextLinkString) ? nextLinkString : string.Empty;
```

src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
andrueastman: We do check before this that the stream is seekable on if block above. Which is the situation when the exception could be thrown. Is there another scenario where we expect the length to throw an error?

andrueastman: The seekable constraint is also due to the need to split up the stream into smaller substreams. So that task can jump to indexes in the stream as it slices it up for upload.  
At the moment, if a stream is not seek able, the task wouldn't be able to slice it up.

baywet: oh right, I forgot about the slicing. Is there a way people could tell us not to slice and just push the whole stream in a single request?

andrueastman: That would be possible by passing a slice size that is the same size as the stream. 
The alternative would be to simply use the request builders that the task uses to make a put a request with the target stream as the payload.

baywet: right, thanks for the additional context and disregard my comment then. I'm guessing it'll be an opportunity for improvement if we receive customer evidence for that. They'll have to buffer the stream in the meantime.

baywet (approved):

PR #636, opened 2023-03-29
feat: New batch for failed requests

svrooij:
- Add `NewBatchWithFailedRequests` to both **BatchRequestContent** and **BatchRequestContentCollection** accepting the dictionary from `batchResponse.GetResponsesStatusCodesAsync()`

### Other links
- Related to https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1759

I don't know where the docs are, but maybe add something like this.

```csharp
var client = new GraphServiceClient(...);
var batch = new BatchResponseContentCollection(client, 20);

// Add steps here with
await batch.AddRequestStepAsync(....);

// Execute batch
var batchResponse = await client.Batch.PostAsync(batch);

// Get status codes
var responseStatuses = await batchResponse.GetResponsesStatusCodesAsync();

// Validate success
var success = responseStatuses.All(s => BatchResponseContent.IsSuccessStatusCode(s.Value));

// Retry failure (this part in introduced by this pr)
var retryBatch = batch.NewBatchWithFailedRequests(responseStatuses);
```

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/636)

### DIFF
--- a/src/Microsoft.Graph.Core/Requests/Content/BatchRequestContent.cs
+++ b/src/Microsoft.Graph.Core/Requests/Content/BatchRequestContent.cs
@@ -5,7 +5,6 @@
 namespace Microsoft.Graph
 {
     using Microsoft.Kiota.Abstractions;
-    using Microsoft.Kiota.Http.HttpClientLibrary;
     using System;
     using System.Collections.Generic;
     using System.IO;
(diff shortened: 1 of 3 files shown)


src/Microsoft.Graph.Core/Requests/Content/BatchRequestContent.cs
```diff
@@ -153,6 +152,25 @@ public bool RemoveBatchRequestStepWithId(string requestId)
             return isRemoved;
         }
 
+        /// <summary>
+        /// Creates a new <see cref="BatchRequestContent"/> with all <see cref="BatchRequestStep"/> that failed.
+        /// </summary>
+        /// <param name="responseStatusCodes">A dictionary with response codes, get with batchResponseContent.GetResponsesStatusCodesAsync()</param>
+        /// <returns>new <see cref="BatchRequestContent"/> with all failed requests.</returns>
+        public BatchRequestContent NewBatchWithFailedRequests(Dictionary<string, HttpStatusCode> responseStatusCodes)
```
andrueastman, line 160: Any chance it makes sense to make `BatchResponseContent.IsSuccessStatusCode` an optional functional parameter that defaults to it so that it can be changed?
I suspect that automatically filtering every status code above 299 may not be enough and a user may want to pass in their own lambda to retry only specific codes they know if what they are looking out for.

svrooij: I totally agree with that we have to do something with that, but I was struggling with what was the best way. With the code in this PR, you can already do something like this before calling that method, if you don't want to retry all faulty status codes.

```csharp
responseStatuses.Where(r => r.Value == (HttpStatusCode)429);
```

My other idea was to use
```csharp
public BatchRequestContent NewBatchWithFailedRequests(Dictionary<string, HttpStatusCode> responseStatusCodes, params HttpStatusCode retryStatuses) {...}
```

But because the library targets `netstandard2.0` the `HttpStatusCode.TooManyRequests` is not available every where.

Open for suggestions, let's pick the best one.

andrueastman (approved):
Thanks for this @svrooij 

Any chance you can also create a PR to improve the docs with the examples at https://github.com/microsoftgraph/msgraph-sdk-dotnet/blob/dev/docs/upgrade-to-v5.md#batch-requests

PR #745, opened 2023-10-20
- enables trimming

baywet:
depends on https://github.com/microsoft/kiota-abstractions-dotnet/pull/139
https://github.com/microsoft/kiota-serialization-json-dotnet/pull/137
https://github.com/microsoft/kiota-serialization-form-dotnet/pull/75
https://github.com/microsoft/kiota-serialization-multipart-dotnet/pull/27
https://github.com/microsoft/kiota-serialization-text-dotnet/pull/108
https://github.com/microsoft/kiota-http-dotnet/pull/155
https://github.com/microsoft/kiota-authentication-azure-dotnet/pull/131
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/745)

### DIFF
--- a/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
+++ b/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj
@@ -59,13 +59,13 @@
     <PackageReference Include="System.Security.Claims" Version="4.3.0" />
     <PackageReference Include="Microsoft.IdentityModel.Protocols.OpenIdConnect" Version="7.0.3" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.Kiota.Abstractions" Version="1.4.0" />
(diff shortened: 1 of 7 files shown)


baywet:
@andrueastman I managed to drive it down to 2 usage of STJ with reflection in DeltaResponseHandler and ResponseHandler.
What I don't understand in there is why would we ever want to return anything else than parsable/why do we need to support JsonElement? is it only for testing purposes?
If we can fix that last part, we're trimmable

andrueastman:
@baywet After taking another look here, I believe you are right.
The calls to STJ for `JsonSerializer.Deserialize<ModelType>` in the DeltaResponseHandler and ResponseHandler should ideally be removed given the already present `IParsable` contract present in the handlers. 
I suspect I was trying to add them for testing purposes so removing them should be valid and any relevant tests refactored to do the job they were running away from. (If it helps, I can resolve these parts)

andrueastman (approved):

Issue #728, opened 2023-09-20
Improve error information when paging using page Iterator


andrueastman:
```c#
            var users = new List<User>();

            var response = await GraphServiceClient.Groups[groupId].TransitiveMembers
                .GraphUser
                .GetAsync(x => x.QueryParameters.Select = new[]
                    { "id", "displayName", "mail", "otherMails", "userPrincipalName", "accountEnabled", "proxyAddresses", "officeLocation" });
            var pageIterator = PageIterator<User, UserCollectionResponse>.CreatePageIterator(
                GraphServiceClient, response, (user) => { users.Add(user); return true; });
            await pageIterator.IterateAsync();

            return users;
```

The problem is that this happens only for one particular group (we use that code on multiple AADs and all works ok, except one particular group within one AAD). I know that the group has members, but I don't actually have access to authorize as the service principal and debug the code (this happens in production). So the problem is, I'm getting an error, but I don't know what it is about, beside it being http status 400.

I was hoping that this would give me the body of the response in the logs (i.e. when doing `.ToString()` on the `ApiException`), but looking at the code, it doesn't seem like it.

_Originally posted by @macwier in https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/2094#issuecomment-1727036825_

kpreddy89:
Hello,

I am still seeing this generic error message "Microsoft.Kiota.Abstractions.ApiException: The server returned an unexpected status code and no error factory is registered for this code: 400. The graph SDK nuget package version being used is 5.18.0

PR #761, opened 2023-11-06
Fixes errors reported when using the PageIterator

andrueastman:
This PR fixes https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/issues/728

Changes include
- Improve error/exceptions thrown when using the pageIterator by allowing one to pass an errorMapping dictionary.
- Refactors serialization tests to use helpers from https://github.com/microsoft/kiota-abstractions-dotnet/pull/141

### DIFF
--- a/src/Microsoft.Graph.Core/Tasks/PageIterator.cs
+++ b/src/Microsoft.Graph.Core/Tasks/PageIterator.cs
@@ -29,6 +29,7 @@ namespace Microsoft.Graph
         private Func<TEntity, bool> _processPageItemCallback;
         private Func<TEntity, Task<bool>> _asyncProcessPageItemCallback;
         private Func<RequestInformation, RequestInformation> _requestConfigurator;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Graph.Core/Tasks/PageIterator.cs
```diff
@@ -95,6 +98,10 @@ public class PageIterator<TEntity, TCollectionPage> where TCollectionPage : IPar
                 _pageItemQueue = new Queue<TEntity>(pageItems),
                 _processPageItemCallback = callback,
                 _requestConfigurator = requestConfigurator,
+                _errorMapping = errorMapping ?? new Dictionary<string, ParsableFactory<IParsable>> {
```
baywet, line 101: pass a string comparer to the dictionary

baywet (changes requested):

PR #776, opened 2023-11-28
Fix batch replays when setting the request body

andrueastman:
This PR is related to https://github.com/microsoft/kiota-http-dotnet/pull/175

It fixes a bug with batch requests where creating a new batch request from failed requests would fail due to incorrect cloning of the request message as the original would be disposed. 

This PR drops the `CloneAsync` HttpRequestMessage extension method as its only used in one place in the batch code and lets the `BatchRequestContent` class only clone the request content which it needs. Tests have also been added to capture this scenario. 

Fixes https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/2166
Fixes https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1850
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/776)


baywet (approved):
Thanks for making the changes

Debro012:
Thank you for all of the work and support in addressing this issue!

PR #887, opened 2024-08-17
feat: Add option to specify requestId manually when calling AddBatchRequestStepAsync()

### LINKED ISSUE #871: Add support for own request Id when making batch requests
### Is your feature request related to a problem? Please describe the problem.

When doing batch calls i sometimes use a custom requestId. For example if i want to request some group information for multiple different groups, i used the groupId as a requestId in the past to simply get the correct response from the body without the need to have to map a generated requestId to a groupId

### Describe the solution you'd like.

The Method `AddBatchRequestStepAsync` inside `BatchRequestContent.cs` is generating the requestId. This method should have an optional requestId paramter which should get passed up so to the `AddBatchRequestStepAsync` of `BatchRequestContentCollection.cs` function.

so basically i would propose to have `AddBatchRequestStepAsync` in `BatchRequestContentCollection.cs` look like this:
```cs
        public Task<string> AddBatchRequestStepAsync(RequestInformation requestInformation, string? requestId = null) //added string? requestId = null
        {
            SetupCurrentRequest();
#pragma warning disable CS0618
            return currentRequest.AddBatchRequestStepAsync(requestInformation, requestId);
#pragma warning restore CS0618
        }
```

and `AddBatchRequestStepAsync` in `BatchRequestContent.cs` look like this:
```cs
public async Task<string> AddBatchRequestStepAsync(RequestInformation requestInformation, string? requestId = null) //added string? requestId = null
{
    if (BatchRequestSteps.Count >= CoreConstants.BatchRequest.MaxNumberOfRequests)
        throw new ArgumentException(string.Format(ErrorConstants.Messages.MaximumValueExceeded, "Number of batch request steps", CoreConstants.BatchRequest.MaxNumberOfRequests));
    // change start
    if(requestId == null){
        requestId = Guid.NewGuid().ToString()
    }
    // change end -> requestId generation replaced so that it only is generated if no custom requestId is provided
    var requestMessage = await RequestAdapter.ConvertToNativeRequestAsync<HttpRequestMessage>(requestInformation);
    BatchRequestStep batchRequestStep = new BatchRequestStep(requestId, requestMessage);
    (BatchRequestSteps as IDictionary<string, BatchRequestStep>)!.Add(batchRequestStep.RequestId, batchRequestStep);
    return requestId;
}
```

This solution would also be backwords compatible, meaning it has no breaking changes because the new `requestId` parameter of `AddBatchRequestStepAsync` is optional

### Additional context?

If not already in development somewehere & it's agreed that this is in scope for the library i can open a PR with the proposed changes from above

MichaMican:
Fixes #871 

- Adding optional parameter `requestId` to `AddBatchRequestStepAsync()` function to allow manual configiguration of requestId on BatchRequests (See reasoning on issue #871 for details)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/887)

### DIFF
--- a/src/Microsoft.Graph.Core/Requests/Content/BatchRequestContent.cs
+++ b/src/Microsoft.Graph.Core/Requests/Content/BatchRequestContent.cs
@@ -137,7 +137,7 @@ namespace Microsoft.Graph
         {
             if (BatchRequestSteps.Count >= CoreConstants.BatchRequest.MaxNumberOfRequests)
                 throw new ArgumentException(string.Format(ErrorConstants.Messages.MaximumValueExceeded, "Number of batch request steps", CoreConstants.BatchRequest.MaxNumberOfRequests));
-            if (requestId == null)
+            if (string.IsNullOrEmpty(requestId))
             {
                 requestId = Guid.NewGuid().ToString();
             }
--- a/src/Microsoft.Graph.Core/Requests/Content/BatchRequestContent.cs
+++ b/src/Microsoft.Graph.Core/Requests/Content/BatchRequestContent.cs
@@ -130,13 +130,17 @@ namespace Microsoft.Graph
         /// Adds a <see cref="RequestInformation"/> to batch request content
         /// </summary>
         /// <param name="requestInformation">A <see cref="RequestInformation"/> to use to build a <see cref="BatchRequestStep"/> to add.</param>
+        /// <param name="requestId">An optional string that will be used as the requestId of the batch request</param>
         /// <returns>The requestId of the  newly created <see cref="BatchRequestStep"/></returns>
         [Obsolete("Please use the BatchRequestContentCollection for making batch requests as it supports handling more than 20 requests and provides a similar API experience.")]
-        public async Task<string> AddBatchRequestStepAsync(RequestInformation requestInformation)
+        public async Task<string> AddBatchRequestStepAsync(RequestInformation requestInformation, string requestId = null)
         {
             if (BatchRequestSteps.Count >= CoreConstants.BatchRequest.MaxNumberOfRequests)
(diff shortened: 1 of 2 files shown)


MichaMican:
@microsoft-github-policy-service agree

src/Microsoft.Graph.Core/Requests/Content/BatchRequestContent.cs
```diff
@@ -130,13 +130,17 @@ public string AddBatchRequestStep(HttpRequestMessage httpRequestMessage)
...
+        /// <param name="requestId">An optional string that will be used as the requestId of the batch request</param>
         /// <returns>The requestId of the  newly created <see cref="BatchRequestStep"/></returns>
         [Obsolete("Please use the BatchRequestContentCollection for making batch requests as it supports handling more than 20 requests and provides a similar API experience.")]
-        public async Task<string> AddBatchRequestStepAsync(RequestInformation requestInformation)
+        public async Task<string> AddBatchRequestStepAsync(RequestInformation requestInformation, string requestId = null)
         {
             if (BatchRequestSteps.Count >= CoreConstants.BatchRequest.MaxNumberOfRequests)
                 throw new ArgumentException(string.Format(ErrorConstants.Messages.MaximumValueExceeded, "Number of batch request steps", CoreConstants.BatchRequest.MaxNumberOfRequests));
-            string requestId = Guid.NewGuid().ToString();
+            if (requestId == null)
```
andrueastman, line 140:
```suggestion
            if (string.IsNullOrEmpty(requestId))
```

andrueastman (approved):
Thanks for the contribution here @MichaMican

PR #892, opened 2024-08-20
Fix/batch steps order

### LINKED ISSUE #891: Batch Request Steps not guaranteed to retain insertion order
### Describe the bug

We store batch requests steps in a [dictionary](https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/main/src/Microsoft.Graph.Core/Requests/Content/BatchRequestContent.cs#L73) which doesn't guarantee insertion order when reading batch request steps.

When using `dependsOn` within OData Batch requests, the API expects that the dependency appears earlier in the array of requests.
Microsoft Graph fails with "Invalid dependency for request Id..."

### Expected behavior

The developer dictates the order of batch request steps & that the order is retained during serialization. In case of an error, developer needs to adjust their order.

### How to reproduce

Add batch request steps with different keys (UUIDs, ints etc) and test that insertion order is retained when calling `GetBatchRequestContentAsync()`

### SDK Version

3.1.17

### Latest version known to work for scenario above?

_No response_

### Known Workarounds

_No response_

### Debug output

<details><summary>Click to expand log</summary>
```

<Log output here>

```
</details>

### Configuration

_No response_

### Other information

_No response_

Ndiritu:
[`System.Collections.Generic.OrderedDictionary<TKey,TValue>`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.ordereddictionary-2?view=net-9.0) would have been the more ideal fix but it's only supported in .NET 9.

[`System.Collections.Specialized.OrderedDictionary`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.specialized.ordereddictionary?view=net-8.0) is supported by older versions but doesn't support generics meaning lots of [casting from `object` to expected types](https://github.com/dotnet/platform-compat/blob/master/docs/DE0006.md)

This PR:
- adds a custom structure to track insertion order of Batch request steps
- updates batch request content tests comparing strings to do so platform agnostically. Tests fail on linux envts due to expected `\r\n`

closes #891 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/892)

### DIFF
--- a/src/Microsoft.Graph.Core/Requests/Content/BatchRequestContent.cs
+++ b/src/Microsoft.Graph.Core/Requests/Content/BatchRequestContent.cs
@@ -17,6 +17,7 @@ namespace Microsoft.Graph
     using System.Text.Json;
     using System.Threading.Tasks;
     using System.Threading;
+    using System.Collections;
 
     /// <summary>
     /// A <see cref="HttpContent"/> implementation to handle json batch requests.
@@ -28,6 +29,8 @@ namespace Microsoft.Graph
         /// A BatchRequestSteps property.
         /// </summary>
         public IReadOnlyDictionary<string, BatchRequestStep> BatchRequestSteps { get; private set; }
+
+        private ArrayList OrderedBatchRequestIds { get; set; }
 
         /// <summary>
         /// The request adapter for sending the batch request
@@ -71,6 +74,7 @@ namespace Microsoft.Graph
             this.Headers.ContentType = new MediaTypeHeaderValue(CoreConstants.MimeTypeNames.Application.Json);
 
             BatchRequestSteps = new Dictionary<string, BatchRequestStep>();
+            OrderedBatchRequestIds = new ArrayList();
 
             foreach (BatchRequestStep requestStep in batchRequestSteps)
             {
@@ -106,6 +110,7 @@ namespace Microsoft.Graph
                 throw new ArgumentException(ErrorConstants.Messages.InvalidDependsOnRequestId);
             }
             (BatchRequestSteps as IDictionary<string, BatchRequestStep>).Add(batchRequestStep.RequestId, batchRequestStep);
+            OrderedBatchRequestIds.Add(batchRequestStep.RequestId);
             return true;
         }
 
@@ -123,6 +128,7 @@ namespace Microsoft.Graph
             string requestId = Guid.NewGuid().ToString();
             BatchRequestStep batchRequestStep = new BatchRequestStep(requestId, httpRequestMessage);
             (BatchRequestSteps as IDictionary<string, BatchRequestStep>).Add(batchRequestStep.RequestId, batchRequestStep);
+            OrderedBatchRequestIds.Add(requestId);
             return requestId;
         }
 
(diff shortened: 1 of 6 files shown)


baywet (dismissed):

src/Microsoft.Graph.Core/Requests/Content/BatchRequestContent.cs
```diff
@@ -29,6 +30,8 @@ public class BatchRequestContent: HttpContent
         /// </summary>
         public IReadOnlyDictionary<string, BatchRequestStep> BatchRequestSteps { get; private set; }
 
+        private ArrayList OrderedBatchRequestIds { get; set; }
```
andrueastman, line 33: I'm abit unsure about having to double track two structures at a time when adding/remove items as it could be a gotcha if the code evolves. 

Maybe we could have an internal implementation of `IDictionary` similar to https://github.com/microsoft/kiota-dotnet/blob/main/src/abstractions/RequestHeaders.cs and override its enumerator

This way, when we get to .NET 9 we can simply swap the class name.

Thoughts?

baywet (approved):

src/Microsoft.Graph.Core/Requests/Content/BatchRequestContentSteps.cs
```diff
@@ -0,0 +1,253 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.Graph;
+
+/// <summary>
+/// Represents a collection of ordered <see cref="BatchRequestStep"/> objects.
+/// </summary>
+public class BatchRequestContentSteps : IReadOnlyDictionary<string, BatchRequestStep>, IDictionary<string, BatchRequestStep>
```
andrueastman, line 10: We just want to use this internally for now. 

```suggestion
internal class BatchRequestContentSteps : IReadOnlyDictionary<string, BatchRequestStep>, IDictionary<string, BatchRequestStep>
```

src/Microsoft.Graph.Core/Requests/Content/BatchRequestContentSteps.cs
```diff
@@ -0,0 +1,253 @@
...
+    /// Gets the number of elements in the collection.
+    /// </summary>
+    public int Count => _requestIds.Count;
+
+    /// <summary>
+    /// Gets a value indicating whether the collection is read-only.
+    /// </summary>
+    public bool IsReadOnly => true;
+
+    IEnumerable<string> IReadOnlyDictionary<string, BatchRequestStep>.Keys => throw new NotImplementedException();
```
andrueastman, line 67:
```suggestion
    IEnumerable<string> IReadOnlyDictionary<string, BatchRequestStep>.Keys => Keys.AsEnumerable();
```

src/Microsoft.Graph.Core/Requests/Content/BatchRequestContentSteps.cs
```diff
@@ -0,0 +1,253 @@
...
+
+    IEnumerable<BatchRequestStep> IReadOnlyDictionary<string, BatchRequestStep>.Values => throw new NotImplementedException();
```
andrueastman, line 69:
```suggestion
    IEnumerable<BatchRequestStep> IReadOnlyDictionary<string, BatchRequestStep>.Values => Values.AsEnumerable();
```

andrueastman (commented):
Thanks for making the changes here.  Just a couple of suggestions. Otherwise, this looks good to me.

andrueastman (approved):
👍🏼

PR #903, opened 2024-08-30
fix: resolve trimming warnings in project

andrueastman:
This PR resolves all trimming warnings in the project 

- Adding relevant `DynamicallyAccessedMembers` attributes where necessarry and Removing calls of `object.GetType()` (as this results in unresolvable trimming warnings)
- Add a trimming validation project and add a validation build step in the pipeline. 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/903)

### DIFF
--- a/src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
+++ b/src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
@@ -6,14 +6,17 @@ namespace Microsoft.Graph
 {
     using System;
     using System.Collections.Generic;
+    using System.ComponentModel;
     using System.IO;
     using System.Net.Http;
+    using System.Text.Json;
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.Graph.Core.Models;
     using Microsoft.Kiota.Abstractions;
     using Microsoft.Kiota.Abstractions.Authentication;
     using Microsoft.Kiota.Abstractions.Serialization;
+    using Microsoft.Kiota.Serialization.Json;
 
     /// <summary>
     /// Task to help with resumable large file uploads.
@@ -40,7 +43,23 @@ namespace Microsoft.Graph
         /// <param name="maxSliceSize">Max size(in bytes) of each slice to be uploaded. Defaults to 5MB. When uploading to OneDrive or SharePoint, this value needs to be a multiple of 320 KiB (327,680 bytes).
         /// If less than 0, default value of 5 MiB is used.</param>
         /// <param name="requestAdapter"><see cref="IRequestAdapter"/> to use for making upload requests. The client should not set Auth headers as upload urls do not need them.</param>
-        public LargeFileUploadTask(IParsable uploadSession, Stream uploadStream, int maxSliceSize = -1, IRequestAdapter requestAdapter = null)
+        [Obsolete("Use the overload that takes in IUploadSession instead.")]
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public LargeFileUploadTask(IParsable uploadSession, Stream uploadStream, int maxSliceSize = -1,
+            IRequestAdapter requestAdapter = null) : this(ExtractSessionFromParsable(uploadSession), uploadStream, maxSliceSize, requestAdapter)
+        {
+        }
+
+        /// <summary>
+        /// Task to help with resumable large file uploads. Generates slices based on <paramref name="uploadSession"/>
+        /// information, and can control uploading of requests.
+        /// </summary>
+        /// <param name="uploadSession">Session information of type <see cref="IUploadSession"/>></param>
+        /// <param name="uploadStream">Readable, seekable stream to be uploaded. Length of session is determined via uploadStream.Length</param>
+        /// <param name="maxSliceSize">Max size(in bytes) of each slice to be uploaded. Defaults to 5MB. When uploading to OneDrive or SharePoint, this value needs to be a multiple of 320 KiB (327,680 bytes).
+        /// If less than 0, default value of 5 MiB is used.</param>
+        /// <param name="requestAdapter"><see cref="IRequestAdapter"/> to use for making upload requests. The client should not set Auth headers as upload urls do not need them.</param>
+        public LargeFileUploadTask(IUploadSession uploadSession, Stream uploadStream, int maxSliceSize = -1, IRequestAdapter requestAdapter = null)
         {
             if (!uploadStream.CanRead || !uploadStream.CanSeek)
             {
@@ -50,8 +69,8 @@ namespace Microsoft.Graph
             {
                 throw new ArgumentException("Must a stream that is not empty");
             }
-            this.Session = ExtractSessionFromParsable(uploadSession);
-            this._requestAdapter = requestAdapter ?? this.InitializeAdapter(Session.UploadUrl);
+            this.Session = uploadSession;
+            this._requestAdapter = requestAdapter ?? InitializeAdapter(Session.UploadUrl);
             this._uploadStream = uploadStream;
             this._rangesRemaining = this.GetRangesRemaining(Session);
             this._maxSliceSize = maxSliceSize < 0 ? DefaultMaxSliceSize : maxSliceSize;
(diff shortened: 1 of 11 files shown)


src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
```diff
@@ -72,14 +74,10 @@ private IUploadSession ExtractSessionFromParsable(IParsable uploadSession)
...
-            var uploadSessionType = uploadSession.GetType();
-
-            return new UploadSession()
-            {
-                ExpirationDateTime = uploadSessionType.GetProperty("ExpirationDateTime").GetValue(uploadSession, null) as DateTimeOffset?,
-                NextExpectedRanges = uploadSessionType.GetProperty("NextExpectedRanges").GetValue(uploadSession, null) as List<string>,
-                UploadUrl = uploadSessionType.GetProperty("UploadUrl").GetValue(uploadSession, null) as string
-            };
+            // convert to local type as we don't have the type info for the upload session just that it implements IParsable
+            var uploadSessionStream = KiotaSerializer.SerializeAsStream(CoreConstants.MimeTypeNames.Application.Json, uploadSession, false);// just in case there's a backing store
```
baywet, line 78: Missing using

src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
```diff
@@ -72,14 +74,10 @@ private IUploadSession ExtractSessionFromParsable(IParsable uploadSession)
...
-
-            return new UploadSession()
-            {
-                ExpirationDateTime = uploadSessionType.GetProperty("ExpirationDateTime").GetValue(uploadSession, null) as DateTimeOffset?,
-                NextExpectedRanges = uploadSessionType.GetProperty("NextExpectedRanges").GetValue(uploadSession, null) as List<string>,
-                UploadUrl = uploadSessionType.GetProperty("UploadUrl").GetValue(uploadSession, null) as string
-            };
+            // convert to local type as we don't have the type info for the upload session just that it implements IParsable
+            var uploadSessionStream = KiotaSerializer.SerializeAsStream(CoreConstants.MimeTypeNames.Application.Json, uploadSession, false);// just in case there's a backing store 
+            var uploadSessionJsonNode = new JsonParseNode(JsonDocument.Parse(uploadSessionStream).RootElement);
```
baywet, line 79: If there is a parse async method, I'm runtime will not yell at people when in aspnet core contexts

src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
andrueastman: not sure what you mean here. Any chance you can confirm?

src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
andrueastman: Just to confirm, do you mean we use `JsonDocument.ParseAsync`?

src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
baywet: stream should be disposed

src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
baywet: yes

andrueastman: As this method is called/used by the constructor, using the `ParseAsync` may not be ideal here. (It will force us to do something like `GetAwaiter.GetResult` as constructors can't be async).
Looking at the implementation of the Parse method, the stream is read into an ReadOnlyMemory/ArraySegment without having to do any task management which seems better creating strings or calling `GetAwaiter.GetResult`.

https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Text.Json/src/System/Text/Json/Document/JsonDocument.Parse.cs#L120C13-L138C10

What do you think?

andrueastman: We did move to using partial classes recently. I will take a look at this and update...

src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
```diff
@@ -63,7 +65,7 @@ public LargeFileUploadTask(IParsable uploadSession, Stream uploadStream, int max
         /// <param name="uploadSession"><see cref="IParsable"/> to initialize an <see cref="IUploadSession"/> from</param>
         /// <returns>A <see cref="IUploadSession"/> instance</returns>
         /// <exception cref="NotImplementedException"></exception>
-        private IUploadSession ExtractSessionFromParsable(IParsable uploadSession)
+        internal static IUploadSession ExtractSessionFromParsable(IParsable uploadSession)
         {
             if (!uploadSession.GetFieldDeserializers().ContainsKey("expirationDateTime"))
```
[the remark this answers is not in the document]
andrueastman: Resolved in latest update.

src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
andrueastman: @baywet The latest updates here will mark the constructor with the IParsable obsolete going forward. 

What this means is that we'll add a new constructor that takes in the `IUploadSession` interface so there won't be any need for serializing/deserializing the members once these two updated to the service library are pulled in.

- https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/pull/896
- https://github.com/microsoftgraph/msgraph-sdk-dotnet/pull/2653

The expectation really is that this path will only be realised in a scenario where a SDK user is using the latest version of the core package without an older version of the SDK models(which won't happen if they are using the core package via the service lib).

The diff is a bit noisy at the moment, so I'll mark this as ready for review once the cleanup work in https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/909 is merged.

src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
```diff
@@ -40,7 +42,22 @@ private IUploadSession Session
         /// <param name="maxSliceSize">Max size(in bytes) of each slice to be uploaded. Defaults to 5MB. When uploading to OneDrive or SharePoint, this value needs to be a multiple of 320 KiB (327,680 bytes).
         /// If less than 0, default value of 5 MiB is used.</param>
         /// <param name="requestAdapter"><see cref="IRequestAdapter"/> to use for making upload requests. The client should not set Auth headers as upload urls do not need them.</param>
-        public LargeFileUploadTask(IParsable uploadSession, Stream uploadStream, int maxSliceSize = -1, IRequestAdapter requestAdapter = null)
+        [Obsolete("Use the overload that takes in IUploadSession instead.")]
```
baywet, line 45:
```suggestion
        [Obsolete("Use the overload that takes in IUploadSession instead.")]
        [EditorBrowsable(EditorBrowsableState.Never)]
```

src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
```diff
@@ -8,12 +8,14 @@ namespace Microsoft.Graph
     using System.Collections.Generic;
     using System.IO;
```
baywet, line 9:
```suggestion
    using System.IO;
    using System.ComponentModel;
```

src/Microsoft.Graph.Core/Tasks/LargeFileUploadTask.cs
baywet: Thanks for looking into this, it's a much better approach!
Since the model will implement both IUploadSession and IParsable, and since both constructors are identical beyond the first argument, I'm a bit worried the compiler might choose the first one and people will need to add casting to get the "right" one. Is this something that you ran into during testing?

andrueastman: Actually, that's sort of what motivated the opening of https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/909.

Once that is closed the diff will get much clearer and the tests should show that the model not using the `IUploadSession ` will lead to the obsolete usage warnings....

baywet (approved):

PR #909, opened 2024-09-04
chore: cleanup usage of obsolete members and async suffixes in test project

andrueastman:
Cleans up obsolete members in test project to make avoid polluting diff in https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/903

Also enable task analyzers in test project.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/909)


tests/Microsoft.Graph.DotnetCore.Core.Test/Requests/BatchRequestBuilderTests.cs
```diff
@@ -2,6 +2,7 @@
 //  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
 // ------------------------------------------------------------------------------
 
+#pragma warning disable CS0618 // Type or member is obsolete
```
baywet, line 5: which member is obsolete? could we move this suppression to just the member that's causing a problem so it does not mask potential future obsolescence?

baywet (changes requested):

tests/Microsoft.Graph.DotnetCore.Core.Test/Requests/BatchRequestBuilderTests.cs
andrueastman: Moved this for this specific file. 
In general the BatchRequests are marked obsolete to encourage use of the BatchCollection types. So, the test files touching those are easier marked for the whole file give that all the APIs are being redirected to the new APIs

baywet (approved):
Thank you for making the changes!

PR #980, opened 2025-09-27
feat: Added statusCodesToTreatAsSuccess (additive success set) to NewBatchWithFailedRequests and Preserving RequestId

jeppesc11:
- Preserving RequestId while using `NewBatchWithFailedRequests `
- Added **statusCodesToTreatAsSuccess** (additive success set) to `NewBatchWithFailedRequests`
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/980)

### DIFF
--- a/src/Microsoft.Graph.Core/Requests/Content/BatchRequestContentCollection.cs
+++ b/src/Microsoft.Graph.Core/Requests/Content/BatchRequestContentCollection.cs
@@ -183,7 +183,13 @@ namespace Microsoft.Graph
             {
                 if (steps.ContainsKey(response.Key) && !BatchResponseContent.IsSuccessStatusCode(response.Value))
                 {
-                    request.AddBatchRequestStep(steps[response.Key].Request);
+                    var step = steps[response.Key];
+                    var newStep = new BatchRequestStep(
+                        requestId: step.RequestId,
+                        httpRequestMessage: step.Request,
+                        dependsOn: step.DependsOn?.ToList());
+
+                    request.AddBatchRequestStep(newStep);
                 }
             }
             return request;
--- a/src/Microsoft.Graph.Core/Requests/Content/BatchRequestContentCollection.cs
+++ b/src/Microsoft.Graph.Core/Requests/Content/BatchRequestContentCollection.cs
@@ -194,5 +194,58 @@ namespace Microsoft.Graph
             }
             return request;
         }
+
+        /// <summary>
+        /// Creates a new <see cref="BatchRequestContentCollection"/> with all <see cref="BatchRequestStep"/> that failed.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


jeppesc11:
@microsoft-github-policy-service agree

jeppesc11:
@ramsessanchez I can see that I should have explained why - apologies.

The reason is that, for example, with “Get file content” you don’t get an ID back that you can use to correlate with your requests -only the batch request ID. So if some number of the 20 calls in the batch fail and I then retry, I no longer have a way to correlate request and response.

I have previously written a longer blog post about this exact issue.
- https://jeppe-spanggaard.dk/blogs/graph-batching-file-content-mapping/
- https://jeppe-spanggaard.dk/blogs/graph-batch-smart-retry/

The reason for statusCodesToTreatAsSuccess is, again, that with “Get File Content” about 9 out of 10 times I get a response code indicating Redirect or Found. That means these will, by default, be retried - even though it’s a valid status code for this scenario. The same applies to 404.

src/Microsoft.Graph.Core/Requests/Content/BatchRequestContentCollection.cs
```diff
@@ -183,10 +183,69 @@ public BatchRequestContentCollection NewBatchWithFailedRequests(Dictionary<strin
...
+        {
+            var request = new BatchRequestContentCollection(this.requestAdapter, batchRequestLimit);
+            if (responseStatusCodes == null || responseStatusCodes.Count == 0)
+            {
+                return request;
+            }
+
+            HashSet<HttpStatusCode> successSet = statusCodesToTreatAsSuccess != null
+                ? [.. statusCodesToTreatAsSuccess]
+                : null;
```
gavinbarron: Why do we have the conversion from IEnumerable to HashSet inside the method? IMO we'd be better served requiring that the HashSet be passed in from the caller.

src/Microsoft.Graph.Core/Requests/Content/BatchRequestContentCollection.cs
```diff
@@ -183,10 +183,69 @@ public BatchRequestContentCollection NewBatchWithFailedRequests(Dictionary<strin
...
+                    return true;
+                }
+
+                if (successSet != null && successSet.Contains(code))
+                {
+                    return true;
+                }
+
+                return false;
+            }
```
gavinbarron:
```suggestion
            bool IsSuccess(HttpStatusCode code) => 
                BatchResponseContent.IsSuccessStatusCode(code) 
                || successSet?.Contains(code) ?? false;
```

gavinbarron (changes requested):

jeppesc11 (commented):
Looks fine with the changes made.

jeppesc11 (commented):
Looks fine

PR #1026, opened 2026-04-17
fix(security): upgrade .NET SDK to 10.0.202

gavinbarron:
## Summary

Update all `global.json` files to use .NET 10.0.202 SDK, replacing outdated .NET 6 and .NET 9 versions that have reached end-of-life.

### Changes
| File | Old Version | New Version |
|------|------------|-------------|
| `global.json` (root) | 6.0.417 | 10.0.202 |
| `src/Microsoft.Graph.Core/global.json` | 6.0.417 | 10.0.202 |
| `tests/.../Trimming/global.json` | 9.0.102 | 10.0.202 |

### Why
.NET 6 reached end-of-life and no longer receives security patches. Upgrading to .NET 10 ensures continued security support.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/1026)

### DIFF
--- a/.github/workflows/validatePullRequest.yml
+++ b/.github/workflows/validatePullRequest.yml
@@ -23,10 +23,15 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
-      - name: Setup .NET
+      - name: Setup .NET 6
         uses: actions/setup-dotnet@v5
         with:
           dotnet-version: 6.0.x
+
+      - name: Setup .NET 10
+        uses: actions/setup-dotnet@v5
+        with:
+          dotnet-version: 10.0.x
 
       - name: Setup JDK for android targets
         uses: actions/setup-java@v5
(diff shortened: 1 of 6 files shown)


.github/workflows/validatePullRequest.yml
```diff
@@ -23,11 +23,16 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
-      - name: Setup .NET
+      - name: Setup .NET 6
```
ramsessanchez, line 26: Are we keeping .net6 here? Or should we be removing this? All subsequent steps use 10

gavinbarron: we need to keep it and have both based on other pipelines like this

ramsessanchez (approved):

PR #1089, opened 2026-08-26
feat: forward request options to Kiota's default handlers

esidorenko-sl:
Related to #1004

- Add a `CreateDefaultHandlers(GraphClientOptions, IRequestOption[])` overload that passes the options through to `KiotaClientFactory.CreateDefaultHandlers`.
- Keep the existing `CreateDefaultHandlers(GraphClientOptions = null)` unchanged, so this is source and binary compatible. It now delegates to the new overload with no options.
- Add a test covering a custom `RetryHandlerOption` reaching the `RetryHandler` that comes back.

Kiota's `CreateDefaultHandlers` already takes an `IRequestOption[]` for configuring the handlers it builds. `GraphClientFactory` calls it with no arguments, so those options are dropped before a caller can reach them.

### Other links
#1004 was closed as by design, and I understand the reasoning: swapping a handler out is a supported approach. I am opening this anyway for two reasons. The overload is additive, so it costs nothing to keep and breaks nothing. And the request in #1004 was to add an optional parameter to the existing method, which would have been a binary breaking change, so it is worth separating the idea from the way it was originally proposed.

If you would still rather not widen the surface here, close it and I will not push back.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/pull/1089)


esidorenko-sl:
@peombwa this has been open since August 26 with no comments yet. I rebased it onto main to clear the BEHIND state, CI is green and the CLA is signed. Locally the unit suite passes on net10.0 and net462, 196 tests each. Could you or someone on the team take a look.

